PR #7, opened 2016-11-13
Add ledger-post-{increment,decrement}-date

awalker4:
Add functions to easily move a post to a different date.

### DIFF
--- a/ledger-post.el
+++ b/ledger-post.el
@@ -194,6 +194,40 @@ regular text."
             (insert "  "))
           (calc))))))
 
+(defun ledger-post-increment-date ()
+  "Increment date of current posting by one day"
+  (interactive)
+  (ledger-post-add-to-date 1))
+
+(defun ledger-post-decrement-date ()
+  "Decrement date of current posting by one day"
+  (interactive)
+  (ledger-post-add-to-date -1))
+
+(defun ledger-post-add-to-date (num)
+  "Add num days to the date of the current posting."
+  (save-excursion
+    (ledger-navigate-beginning-of-xact)
+    (let* ((beg (point))
+           (end (re-search-forward ledger-iso-date-regexp))
+           (xact-date (filter-buffer-substring beg end)))
+      (delete-region beg end)
+      (insert
+       (format-time-string
+        (or (cdr (assoc "date-format" ledger-environment-alist))
+            ledger-default-date-format)
+        (time-add (ledger-encoded-date xact-date)
+                  (days-to-time num)))))))
+
+(defun ledger-encoded-date (date)
+  "Given a date in the form described by ledger-iso-date-regexp,
+   return encoded time string"
+  (string-match ledger-iso-date-regexp date)
+  (encode-time 0 0 0
+               (string-to-number (match-string 4 date))
+               (string-to-number (match-string 3 date))
+               (string-to-number (match-string 2 date))))
+
 (provide 'ledger-post)
 
 
--- a/ledger-mode.el
+++ b/ledger-mode.el
@@ -310,6 +310,8 @@ With a prefix argument, remove the effective date."
     ["Delete Transaction" ledger-delete-current-transaction]
     "---"
     ["Calc on Amount" ledger-post-edit-amount]
+    ["Increment xact date" ledger-post-increment-date]
+    ["Decrement xact date" ledger-post-decrement-date]
     "---"
     ["Check Balance" ledger-display-balance-at-point ledger-works]
     ["Reconcile Account" ledger-reconcile ledger-works]


awalker4:
Hey everyone! I'm still a bit unfamiliar with the code so I have a few questions.

- Is ledger-post the right file for these?
- How would you suggest I add a test for this?
- Can anyone recommend a keybinding? I was think C-M-. and C-M-,

Thanks!

ledger-post.el
```diff
@@ -194,6 +194,40 @@ regular text."
...
+  "Decrement date of current posting by one day"
+  (interactive)
+  (ledger-post-add-to-date -1))
+
+(defun ledger-post-add-to-date (num)
+  "Add num days to the date of the current posting."
+  (save-excursion
+    (ledger-navigate-beginning-of-xact)
+    (let* ((beg (point))
+           (end (re-search-forward ledger-iso-date-regexp))
```
purcell: So this only works with ISO dates? That seems suboptimal, even if I personally think everyone should be using that format.

purcell:
> Is ledger-post the right file for these?

Yes.

> Can anyone recommend a keybinding? I was think C-M-. and C-M-,

Those are a bit clunky IMO, and a bit too reminiscent of the tag bindings (`M-.` & `M-,`)

IIRC org has similar functionality, so maybe copy its bindings?

thdox:
Some time ago, I was thinking about the same kind of functionality, that is increment/decrement date.
But I was thinking to implement it (I have nothing implemented yet), by re-using the way it works in org-mode.
When shift-up or shift-down on date, and wherever is the cursor (year or month or day), the figure is incremented/decremented. The way it works in org is conflicting with shift selection (this is documented). Also, in org-mode, it is working using the context, that is, it only work when the cursor is *on* a date, enabling to use shift-up and shift-down elsewhere.

purcell:
I think the org date format is more constrained than the ledger format, so it might not be easy/possible to do the same.

enderw88:
I think this is a good idea, but I would prefer it to work only when point is within a date already. Either that or return point to where you were after executing the date change.  We get lots of complaints when we do things that move point when not absolutely necessary.

ledger-post.el
awalker4: Ah, that was my other question! Does ledger-mode support other formats?

awalker4:
I was originally inspired by the org way of doing dates, so I'll definitely have to see if I can repurpose those. Thanks everyone!

enderw88:
Ledger-mode is agnostic to the date format except for the sort functions. I
have had a few requests to expand the capabilities, but haven't had the
energ,y it is non trivial.

thdox:
To help with suggestion from @enderw88 

> I would prefer it to work only when point is within a date already

 I would suggest to use `ledger-context-at-point` instead of `(re-search-forward ledger-iso-date-regexp)`.

cpitclaudel:
I'd like to see this feature merged :) I'll add a few comments.
Regarding the binding: Org's binding is too generic indeed, but we could use something like M-S-up and M-S-down.  In Org these move lines around, but on clock entries they act as S-up, which increments times.

ledger-post.el
cpitclaudel: Ledger does support other formats, and ledger-mode doesn't care (except for ledger-xact, where it assumes iso:

```elisp
      (let* ((date (car args))
             (parsed-date (ledger-parse-iso-date date)))
        (setq ledger-add-transaction-last-date parsed-date)
        (push-mark)
        ;; TODO: what about when it can't be parsed?
        (ledger-xact-find-slot (or parsed-date date))))
```

The problem is that Emacs doesn't have strptime, so there isn't much we could do (short of calling ledger) to figure out the date.

We might want to warn if the current user-specified date format doesn't match the ones recognized by ledger-iso-date-regexp.

ledger-post.el
```diff
@@ -194,6 +194,40 @@ regular text."
...
+           (xact-date (filter-buffer-substring beg end)))
+      (delete-region beg end)
+      (insert
+       (format-time-string
+        (or (cdr (assoc "date-format" ledger-environment-alist))
+            ledger-default-date-format)
+        (time-add (ledger-encoded-date xact-date)
+                  (days-to-time num)))))))
+
+(defun ledger-encoded-date (date)
```
cpitclaudel: I think this is `ledger-parse-iso-date` in ledger-xact, so no need to duplicate it here :)

ledger-post.el
```diff
@@ -194,6 +194,40 @@ regular text."
             (insert "  "))
           (calc))))))
 
+(defun ledger-post-increment-date ()
+  "Increment date of current posting by one day"
```
cpitclaudel: Missing "."

ledger-post.el
```diff
@@ -194,6 +194,40 @@ regular text."
...
+  (interactive)
+  (ledger-post-add-to-date 1))
+
+(defun ledger-post-decrement-date ()
+  "Decrement date of current posting by one day"
+  (interactive)
+  (ledger-post-add-to-date -1))
```
cpitclaudel: Missing "."

cpitclaudel:
I wonder if it'd be worth also moving the entry to the proper location (ledger-xact-find-slot).  @awalker4, are you still interested in getting this merged? :)

DamienCassou:
Should this PR be closed?

PR #9, opened 2016-12-02
Makes ledger-report-mode use the header-line for its preface

atheriel:
At present `ledger-report-do` will insert the report name, command, and a separator line of `===` characters into the buffer as a preface to the output of the report command itself. This seems like functionality that belongs in Emacs's `header-line-format` facility instead.

This PR adds a demo header line containing the report name, buffer, and command as a replacement for the current preface approach.

Sample output before:

screen shot 1

After:

screen shot 2

The current content and faces of the header are somewhat arbitrary, and I'm open to suggestions as to how to improve them.

### DIFF
--- a/ledger-report.el
+++ b/ledger-report.el
@@ -88,6 +88,18 @@ reports to their location in the currrent ledger file buffer."
   :type 'boolean
   :group 'ledger-report)
 
+(defcustom ledger-report-use-header-line nil
+  "When non-nil, indicate the report name and command in the `header-line'
+instead of in the buffer."
+  :type 'boolean
+  :group 'ledger-report)
+
+(defcustom ledger-report-header-line-fn #'ledger-report--header-function
+  "When `ledger-report-use-header-line' is non-nil, evaluate this function
+in the `header-line'."
+  :type 'function
+  :group 'ledger-report)
+
 (defvar ledger-report-buffer-name "*Ledger Report*")
 
 (defvar ledger-report-name nil)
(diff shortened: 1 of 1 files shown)


enderw88:
Not sure I really like this.  Seeing the complete command that was run is more important than knowing what buffer it originated from.

atheriel:
I suppose that I usually hit `e` to see what the command was, since they're often longer than the width of the frame anyway. But I can see what you mean.

Would it be enough to nix the buffer name in the middle, or are you against using the `header-line` altogether?

atheriel:
Example:

screen shot 3

This is about the same width as the current command output.

enderw88:
I would like to hear from other users.  If no one else pipes in by Sunday I will merge it in as is.

purcell:
I find the "before" version easier to read and less magical, to be honest, but there's a slight benefit to the contents of the buffer containing only the output of the command. I'm therefore just slightly -1 overall.

thdox:
From my point of view, this is just a change of color and position of information. As such this is something like "there's no accounting for taste" or "your mileage may vary". My personal taste would slightly favor "before" state, because from my point of view there is more "space" and this is slightly more readable.

enderw88:
Please add this as an option which can be configure by the user.  Then I will merge it.

atheriel:
@enderw88 That's a good compromise, although the lack of enthusiasm in this thread makes me think that few users will actually make use of this. It seems a bit silly to merge a patch just to make one user (me) happy, so I'm open to this remaining out of tree.

However, I did implement your suggestion. To make things more flexible (and perhaps encourage usage), I've made it possible to set the function that will be evaluated in the header line, as well as gating the feature itself behind a switch.

I've rebased on this branch, so the patch is still contained in a single commit.

PR #24, opened 2017-02-07
Fix ledger string to number

felipebalbi:
With these two patches, I have reconciliation working with decimal comma.

### DIFF
--- a/ledger-commodities.el
+++ b/ledger-commodities.el
@@ -81,9 +81,9 @@ Returns a list with (value commodity)."
            ((re-search-forward "0" nil t)
             ;; couldn't find a decimal number, look for a single 0,
             ;; indicating account with zero balance
-            (list 0 ledger-reconcile-default-commodity))))
-      ;; nothing found, return 0
-      (list 0 ledger-reconcile-default-commodity))))
+            (list 0 ledger-reconcile-default-commodity))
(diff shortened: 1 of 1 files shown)


enderw88:
I need to understand the purpose of the [] in the expression.  I have tested the previous code with decimal-comma set and it worked.  Was it not working for you?

felipebalbi:
@enderw88 [.] is similar to \. I'm telling regexp to treat . as a character, not a field that matches any single character. As described in the commit log, I had every character removed from the string and always got "pending" set to 0. Everything worked fine when I reversed all numbers to decimal period, though.

purcell:
> [.] is similar to . I'm telling regexp to treat . as a character, not a field that matches any single character

You should ideall use `"\\."` instead here: the `\\` in a string expands to a literal `\`, which escapes the following special regexp character, `.` in this case.

thdox:
In this pull request, you say that "With these two patches, I have reconciliation working with decimal comma". 

I live in France, the decimal separator is comma. Reconciliation is working fine for my use cases. Moreover, in the regression tests, there are the tests `ledger-reconcile/test-008` and `ledger-reconcile/test-012` that are passing with comma as decimal separator.

Do you have an example of a ledger file that is failing when reconciliation is launched?

PR #26, opened 2017-03-14
Keep test/ directory of the load-path

tarsius:
Because the files, features, and symbols in "test/" are not prefixed
with "ledger-", we should make it as unlikely as possible for "test/"
to end up on the `load-path`.

By adding the file "test/.nosearch" at least
`normal-top-level-add-to-load-path` and the tools used to maintain the
Emacsmirror are told to ignore test.  Homegrown package managers might
also benefit.

In "test/Makefile", explicitly add "." to the `load-path` to counter
this change in the one case where we want these files to be on the
`load-path`.


thdox:
@tarsius:

About the search patch, I was not aware of this kind of issue, and I recognize that impacts may exist, and the dot file sounds to be a solution.

About the add of "-L .", I do not understand, as for me the _short_ option "-L ." is the same as the _long_ option "--directory .", and this long option is already present in Makefile.

Cheers

tarsius:
> this long option is already present in Makefile.

I've overlooked that. It's fixed now.

thdox:
Thanks

tarsius:
Thanks to you too!

PR #29, opened 2017-04-10
make comment-start buffer-local

csrhodes:
otherwise fill-paragraph in text mode does wrong things if, like me, you
use semicolons a lot in prose.


purcell:
Good catch, thanks.

PR #30, opened 2017-04-14
Use cl-lib package rather than cl

purcell:
Switching to `cl-lib` has been a to-do for a while. This PR makes the change, and also switches the travis build to evm and cask, which make cross-Emacs testing and package installation easy.

NOTE: I don't expect the travis build for this branch to work immediately, so this PR should be considered a work in progress.

### DIFF
--- a/ledger-check.el
+++ b/ledger-check.el
@@ -28,8 +28,8 @@
 
 (require 'easymenu)
 (require 'ledger-navigate)
-(eval-when-compile
-  (require 'cl))
+(require 'ledger-report) ; for ledger-master-file
+
 
 (defvar ledger-check-buffer-name "*Ledger Check*")
 (defvar ledger-original-window-cfg nil)
--- a/ledger-context.el
+++ b/ledger-context.el
@@ -25,9 +25,6 @@
 
 ;;; Code:
 
-(eval-when-compile
-  (require 'cl))
-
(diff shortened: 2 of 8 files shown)


enderw88:
Why is cl-lib preferred?

purcell:
`cl` is now deprecated wholesale. A little of the history is covered in the [manual page](https://www.gnu.org/software/emacs/manual/html_mono/cl.html).

For the most part, `cl-lib` is a drop-in replacement, with a consistent `cl-` prefix for identifiers, and it is built in from Emacs 24.3 onward. For earlier Emacsen, or where the built-in version is not new enough, there's a `cl-lib` package in GNU ELPA. This means there's really no reason to stick with `cl`.

(The current `master` Travis build only checks Emacs 24.3 and great anyway, so we should probably declare that as the minimum supported Emacs version for `ledger-mode`.)

purcell:
I've backed out the Travis build changes, which leaves just the `cl` => `cl-lib` changes in this PR.

The build is passing, or at least as much it was previously passing: the failure in `context-test` with Emacs 24.3 and `ledger` "master" has been around since we merged #18 from @thdox a while ago. :-/

purcell:
Going ahead and merging this, since the build failure is the same as in `master`.

PR #35, opened 2017-05-31
Add a section "Basic usage" to README

Kaligule:
This would have helped me when I started using ledger mode.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,15 @@ Related packages
 In-buffer checking of formatting and balancing of transactions is
 available using [`flycheck-ledger`][flycheck-ledger].
 
+Basic usage
+===========
+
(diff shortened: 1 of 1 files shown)


purcell:
Thanks, but while those are a few hints that would have helped you specifically, it's not close to an adequate "basic usage" guide. I'd expect new users to use the usual `F1 m` trick to discover commands, and to browse options using the `customize` interface. Writing a full guide would be a much bigger endeavour, which I think none of us has time to tackle, sorry.

purcell:
I added a little note with the above tips - thanks!

Kaligule:
Ok no problem. Today I learned about `C-h m`. Thanks.

PR #48, opened 2017-08-04
Implement date completion

yurikhan:
David Glasser asked on the mailing list for a faster way to enter transaction dates, to which I suggested this idea:

> I think I might like Tab to autocomplete and canonicalize dates. The
idea is that if you typed a single number at the beginning of line and
pressed Tab, it would expand to the date in the current or last month;
two numbers with a separator would expand to the date in the current
or last year; and separators would be normalized according to
`ledger-default-date-format`. (The “current or last” in both cases would
resolve to yield a date that is not in the future.)
>
> Example: On 2018-01-03, you enter `30` at the start of line, press
Tab, and it expands to `2017-12-30`.

This patch is my attempt to implement that.

### DIFF
--- a/ledger-complete.el
+++ b/ledger-complete.el
@@ -23,6 +23,7 @@
 ;; Functions providing payee and account auto complete.
 
 (require 'pcomplete)
+(require 'cl-lib)
 
 ;; In-place completion support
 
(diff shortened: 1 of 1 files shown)


purcell:
Looks good to me.

glasser:
Is the test failure real or spurious? I'd love to see this merged.

glasser:
Though one note after using it: it might be nice if after completing the date, it added a space after it?

yurikhan:
Good idea about the space, and I will look into the test breakage.

purcell:
Merged. The build failure is spurious, and has been happening on one specific Emacs version for a while. :-/

yurikhan:
Hm, I was going to add a space after the date but that will interfere with effective dates. I don’t use effective dates myself; are they usually entered at the same time as the main date, or added afterwards? Come to think of it, should completion be also supported for the effective date?

glasser:
I don't use effective dates either. I suppose this could be a custom option.

glasser:
Note: I think I want the extra space if I'm looking at the end of the line but not if there's already a space there (eg I'm expanding a date I forgot to expand previously)

yurikhan:
I added two commits to the branch. One adds a space after the completed date if at the end of line. The other adds completion for effective dates.

While the transaction date resolves to the latest date is before the current date, the effective date resolves to the earliest date that is after the transaction date. Example: `2017-08-31=02` `TAB` → `2017-08-31=2017-09-02`.

No option for the space yet. For now, I’m inclined to think effective dates are not entered initially but added later.

https://github.com/ledger/ledger-mode/compare/master...yurikhan:yk/date-completion

Should I open a new pull request?

purcell:
> Should I open a new pull request?

Yes please!

yurikhan:
#50

PR #55, opened 2017-11-15
Initial implementation of "see as you type" transaction creation

jwiegley:
@enderw88 Here's an initial implementation that shows what I mean. It's not quite ready for prime-time for the following reasons:

 1. Error handling is terrible. It just swallows chance errors so as not to disrupt the display.

 2. It's slow, because it copies over the source buffer for literally every key pressed. But this is to take advantage of "prior knowledge" in what gets displayed, so in a way it's necessary.

 3. It feels fragile overall, and I'm not sure my window management represents best practices.

However, it does work very nicely (for small ledger files at least), showing a colorized live display of what will get inserted.

### DIFF
--- a/ledger-xact.el
+++ b/ledger-xact.el
@@ -195,6 +195,90 @@ correct chronological place in the buffer."
         (insert (car args) " \n\n")
         (end-of-line -1)))))
 
+(defvar ledger--source-buffer)
+
+(defun ledger--get-transaction (transaction-text)
+  "Generate a transaction from TRANSACTION-TEXT and insert it at
+the current buffer position."
+  (let* ((args (with-temp-buffer
+                 (insert transaction-text)
+                 (eshell-parse-arguments (point-min) (point-max)))))
+    (if args
+        (if (> (length args) 1)
+            (save-excursion
+              (let ((input
+                     (with-current-buffer
+                         (generate-new-buffer " *Ledger xact input*")
+                       (insert-buffer ledger--source-buffer)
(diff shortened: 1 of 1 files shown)


jwiegley:
Oh, and it needs to incorporate the "figure out where in the timeline this should go" logic.

cpitclaudel:
This works really nicely, actually.  And it doesn't feel that slow here.
Have you had time to look at it again since you wrote the initial implementation?

cpitclaudel:
> It's slow, because it copies over the source buffer for literally every key pressed. But this is to take advantage of "prior knowledge" in what gets displayed, so in a way it's necessary.

Presumably the rest of the buffer should be frozen while the user builds their transaction, so is it actually necessary to resend the whole buffer every time?  Isn't it enough to save a copy at the beginning, and then reuse that?

cpitclaudel:
> Error handling is terrible. It just swallows chance errors so as not to disrupt the display.

I did run into a few issues:

* Ledger runs assertions from the file while the transaction is being typed, so if ledger infers a transaction that doesn't pass assertions the live window just shows 'assertion failed'

* It's tricky to pass special characters to ledger, like `'` or `;`, as they tend to crash ledger-mode. (This is a ledger-mode issue, I think; it's not specific to this patch)

jwiegley:
@cpitclaudel Great point about only needing the file up to the point of insertion. And no, I haven't looked at this code any further since writing it.

PR #59, opened 2017-12-26
Fix `ledger-context-at-point' and balance assertions

hexagonal-sun:
Suppose we have the following posting:

```
2017/12/04 * foobar
    Assets:Reimbursements:barbaz               £-20.22 = £150.69
    Assets:Bank:Checking
```

When we execute `ledger-context-at-point` on the second line which contains the balance assertion we get:

```
(acct-transaction account ((indent "    " 3543)(account "Assets:Reimbursements:barbaz               £-20.22 = £150.69" 3549)))
```

This causes prompts of account names to fail as the amount and balance assertion is included in the account name. 

This patch fixes the issue by including balance assertions in `ledger-line-config`.


purcell:
Thanks. I guess in theory there could also be a comment after the balance assertion, in which case an additional entry might be desirable. I'll merge this for now in any case: the build failure was already in `master`.

PR #60, opened 2017-12-27
Fix group matching of `ledger-commoditized-amount-regexp`

hexagonal-sun:
When `ledger-context-at-point` is applied to:

```
    Expenses:Food:Dining                      £50.00  ;foobar
```

We obtain:

```
(acct-transaction account ((indent "   " 8664) (status nil nil) (account "Expenses:Food:Dining" 8668) (separator "                      " 8688) (commoditized-amount "£50.00" 8710) (separator "£" 8710) (comment "50.00" 8711)))
```

Notice that the match values for separator and comment are the sub-matches of commoditized-amount.  This patch creates no-group variants of the amount and commodity regexps so that they can be used in the commoditized amount regexp.  After this patch the match values are correct:

```
(acct-transaction account ((indent "   " 8664) (status nil nil) (account "Expenses:Food:Dining" 8668) (separator "                      " 8688) (commoditized-amount "£50.00" 8710) (separator "  " 8716) (comment "foobar" 8719)))
```

### DIFF
--- a/ledger-regex.el
+++ b/ledger-regex.el
@@ -275,33 +275,41 @@
                       (kind account-kind)
                       (name account))
 
+(ledger-define-regexp commodity-no-group
+                      (rx (or (and ?\" (+ (not (any ?\"))) ?\")
+                              (not (any blank ?\n
+                                        digit
+                                        ?- ?\[ ?\]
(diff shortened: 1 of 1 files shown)


purcell:
Thanks.

Issue #20, opened 2017-02-04
Request: color code debits and credit differently


jmrgithub:
I'd like to request that debits and credits in the reconcile window be colorized differently, so it's easy to distinguish them.

enderw88:
This is an interesting idea.  i will look into it.

akater:
Debits and credits are color coded by ledger itself. I managed to propagate those colors into emacs but it required an unpackaged elisp source ftp://download.tuxfamily.org/user42/tty-format.el I'll submit a PR when I figure out how to include `tty-format.el` (or the relevant bits) properly.

akater:
I was wrong: `tty-format.el` is actually available as package, from the repository `("user42" . "https://download.tuxfamily.org/user42/elpa/packages/")`. However, as of now, `package.el` has issues with the signature on `tty-format.el`, I'm looking into it.

I'd appreciate if someone told me how to specify tty-format as dependency properly. Do I simply `(require tty-format)` in `ledger-mode.el`?

purcell:
> I'd appreciate if someone told me how to specify tty-format as dependency properly. Do I simply (require 'tty-format) in ledger-mode.el?

That's how you'd do it, yes, together with a `Package-Requires` header entry, but if the package isn't available via a major repository, it's not something we'd want to add as a dependency of `ledger-mode`.

akater:
if the package isn't available via a major repository, it's
>not something we'd want to add as a dependency of `ledger-mode`.

Pity. I will talk to the tty-format developer, and suggest relocation or mirroring.

Straightforward (“static”) inclusion of el files (one file in this case) in not welcome either?

purcell:
> Pity. I will talk to the tty-format developer, and suggest relocation or mirroring.

Yeah, I'd be happy to host it in MELPA, but it would need a couple of little fixes first (primarily because it introduces top-level symbols that aren't prefixed with the package name).

> Straightforward (“static”) inclusion of el files (one file in this case) in not welcome either?

No, because then we'd have to adopt and maintain the code, and also prefix it with `ledger-` etc.

akater:
No need for tty-format dependency. Following a suggestion by tty-format developer Kevin Ryde, I have successfully tested a more straightforward way to do it. Once we're done with #61, I'll submit a patch. (it requires inserting some ledger options, as in #61)

akater:
[...]

PR #61, opened 2017-12-29
choose report width according to window-text-width

akater:
Register reports need as much width as possible, at least on smaller displays.

Notes:
- I replace end-of-line because command string may have multiple lines.
- Console emacs adjustment was chosen empirically.

Side notes:
- I don't use any margins or fringes in emacs. Further tweaks might be needed due to this.
- I guess I use longer account names than usual. The patch was motivated by this.

### DIFF
--- a/ledger-report.el
+++ b/ledger-report.el
@@ -370,7 +370,7 @@ Optional EDIT the command."
      (replace-regexp-in-string "$"
 			       (concat
 				" --columns " (int-to-string
-					       (- (window-width)
+					       (- (window-text-width)
 						  (if (display-graphic-p) 0 2))))
       (if (and register-report
 	       ledger-report-links-in-register
--- a/ledger-report.el
+++ b/ledger-report.el
@@ -367,8 +367,9 @@ Optional EDIT the command."
     (shell-command
      ;; --subtotal does not produce identifiable transactions, so don't
      ;; prepend location information for them
-     (replace-regexp-in-string "$"
-			       (concat
+     (replace-regexp-in-string (concat "^\\([:blank:]*" ledger-binary-path "\\)")
+                               ;; will fail if ledger-binary-path contains whitespace
+			       (concat "\\1"
 				" --columns " (int-to-string
 					       (- (window-text-width)
 						  (if (display-graphic-p) 0 2))))
--- a/ledger-report.el
+++ b/ledger-report.el
@@ -356,49 +356,50 @@ Optional EDIT the command."
   (goto-char (point-min))
   (setq header-line-format (when ledger-report-use-header-line
                              '(:eval (funcall ledger-report-header-line-fn))))
-  (unless ledger-report-use-header-line
-    (insert (format "Report: %s\n" ledger-report-name)
-            (format "Command: %s\n" cmd)
-            (make-string (- (window-width) 1) ?=)
-            "\n\n"))
-  (let ((data-pos (point))
-        (register-report (string-match " reg\\(ister\\)? " cmd))
-        files-in-report)
-    (shell-command
-     ;; --subtotal does not produce identifiable transactions, so don't
-     ;; prepend location information for them
-     (replace-regexp-in-string (concat "^\\([:blank:]*" ledger-binary-path "\\)")
-                               ;; will fail if ledger-binary-path contains whitespace
-			       (concat "\\1"
-				" --columns " (int-to-string
-					       (- (if (fboundp 'window-text-width)
-                                                      (window-text-width)
-                                                      (window-width))
-						  (if (display-graphic-p) 0 2))))
-      (if (and register-report
-	       ledger-report-links-in-register
-	       (not (string-match "--subtotal" cmd)))
-	  (concat cmd " --prepend-format='%(filename):%(beg_line):'")
-	cmd))
-     t nil)
-    (when (and register-report ledger-report-links-in-register)
-      (goto-char data-pos)
-      (while (re-search-forward "^\\(/[^:]+\\)?:\\([0-9]+\\)?:" nil t)
-        (let ((file (match-string 1))
-              (line (string-to-number (match-string 2))))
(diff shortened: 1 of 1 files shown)


purcell:
There's `window-text-width`, in recent Emacsen at least, so no guesses should be necessary.

akater:
I tried to set `--columns` to `(window-text-width)` without further
adjustment, and the report output in tty emacs still looks ugly. (BTW,
in my case window is split horizontally, i.e. reports appear on the
right.) But yes, `window-text-width` is more appropriate.

purcell:
Thanks. Please don't `[ci skip]` all these changes - build feedback is important if code is to be merged.

jwiegley:
Agreed.

akater:
It would be more appropriate to add `--comment` after each `%(binary)` but the latter string is itself varying. Maybe I'll do this tomorrow, maybe in a week.

akater:
The mechanism in its current state will fail if provided path to ledger binary contains whitespace, which is pointed out in the source. I don't know how to implement an alternative that would work around that. Overall, an operation that modifies command line can never be completely fail-safe, and I don't find this particular issue important.

As of Travis check, I don't understand what to do about it. The first error line in the log is,
`E: Failed to fetch http://ppa.launchpad.net/boost-latest/ppa/ubuntu/dists/trusty/main/binary-amd64/Packages  404  Not Found`
I don't think I have any control over this. My patch is unrelated and trivial.

purcell:
> As of Travis check, I don't understand what to do about it. The first error line in the log is,
> E: Failed to fetch http://ppa.launchpad.net/boost-latest/ppa/ubuntu/dists/trusty/main/binary-amd64/Packages 404 Not Found
> I don't think I have any control over this. My patch is unrelated and trivial.

Yes, that failure in the build is our problem to address, but it's still preferable for the build to run.

Regarding the code, I wasn't quite clear why the regex replacement based on the binary path was necessary. It seems like the following (shorter) snippet might do the same thing:
```el
 (let ((data-pos (point))
        (register-report (string-match " reg\\(ister\\)? " cmd))
        files-in-report
        columns (- (window-width) (if (display-graphic-p) 0 2)))
    (shell-command
     (concat cmd " --columns " (int-to-string columns)
             (when (and register-report
                        ledger-report-links-in-register
                        ;; --subtotal does not produce identifiable transactions, so don't
                        ;; prepend location information for them
                        (not (string-match "--subtotal" cmd)))
               " --prepend-format='%(filename):%(beg_line):'"))
     t nil)
```

It also seems like `window-text-width` is very recent, so we'd need to use `window-width` instead, as per the above snippet.

Another potential issue is that the manual notes that `--columns` affects the register report only, so I'm wondering if it's safe to include that argument when the report is not a register report.

Can you try the above out please?

akater:
>Regarding the code, I wasn't quite clear why the regex replacement
>based on the binary path was necessary. 

There might be multiple calls to ledger in a single “command”. E.g., I am using such a (balance) report. One call with a specific `-X` option, another one without.

>
>It also seems like `window-text-width` is very recent, so we'd need to
>use `window-width` instead, as per the above snippet.

How about `fboundp` check for `window-text-width`?

>Another potential issue is that the manual notes that `--columns`
>affects the register report only, so I'm wondering if it's safe to
>include that argument when the report is not a register report.

I tried all default reports in ledger-mode, with no issues.

akater:
Tested with `window-width` and `window-text-width`, with fringes and with scrollbars. Results looked fine in both tty emacs and GUI emacs.

akater:
The same (conceptually) number is utilized above in `(make-string (- window-width 1) ?=)`. Don't know why subtract 1 and not 0 but brought the introduced local variable into accordance with the choice made before.

akater:
Sorry to bother. Just to make it clear, are there any unresolved problems with this PR?

Summary of issues:
#### why regex replacement
Consider the following ledger-report entry:

```
("cash-and-rides" "%(binary) -f %(ledger-file) reg cash
%(binary) -f %(ledger-file) -X rides reg travelcard")
```
Appending options to the tail of the string would apply said options to the second command only, and the output would look ugly. One can also come up with examples of useful commands which don't make use of ledger at all in their last lines.

#### window-text-width is recent
I added an `fboundp` check. Didn't test with Emacs other than 25.3. The current version already uses `window-width` for the same purpose (output formatting), and I adjusted the tweaks accordingly, so the patch conforms to what's already there.

#### --columns option for non-register report
Neither Balance report nor Cleared report choke on `--columns`. I must admit though, I don't use Cleared report at all.

If I need to improve the patch further, please tell. I'd like to treat #20 in a similar way (pre-packaged command line options, maybe activated by a `defcustom`ed switch).

jabranham:
This works well for me. Thanks, @akater! Looking forward to seeing your implementation of colored credits/debits!

cpitclaudel:
The width adjustment part looks nice; thanks! I have an alternate fix for appending parts of the command name.

I don't think we need `window-text-width`; window-width is just fine, but we do indeed need to subtract 1 to account for the continuation glyph column:

```
(window-width &optional WINDOW PIXELWISE)

Return the width of WINDOW’s text area.
WINDOW must be a live window and defaults to the selected one.  Optional
argument PIXELWISE non-nil means return the width in pixels.  The return
value does not include any vertical dividers, fringes or marginal areas,
or scroll bars.

If PIXELWISE is nil, return the largest integer smaller than WINDOW’s
pixel width divided by the character width of WINDOW’s frame.  This
means that if a column at the right of the text area is only partially
visible, that column is not counted.

Note that the returned value includes the column reserved for the
continuation glyph.
```

akater:
I am closing this pull request and recommend https://github.com/cpitclaudel/ledger-mode/tree/report-improvements as a replacement which covers ≈everything suggested here (and more), in a more robust way. The only thing that bothers me is the tty case. There might be something weird about my Emacs setup which necessitated using `(display-graphic-p)` to calculate window width.

PR #68, opened 2018-01-21
Fix report face description & rename to ledger-font-report-face

jabranham:
This makes the face used in report buffers inherit from `ledger-font-default-face` rather than take some weird orange-y color.


purcell:
+1 for inheriting a face instead of using a special case, but why rename it too? It obscures the intent here.

jabranham:
Because, as far as I can tell, it's not actually clickable. Maybe that's a bug though and it should be?

purcell:
Looks good, thanks!

PR #69, opened 2018-01-21
Advertise the info manual on the README

jabranham:


purcell:
Great, thanks.

cpitclaudel:
Might be worth linking to the PDF, too: https://www.ledger-cli.org/3.0/doc/ledger-mode.pdf Or is that not updated regularly?

PR #70, opened 2018-01-22
Drop text properties when gathering account names

cpitclaudel:
This makes the output of ledger-accounts-list-in-buffer and ledger-accounts-tree more readable, which helps when debugging.


purcell:
Makes sense, thanks.

PR #72, opened 2018-01-22
Only highlight current transaction if region is inactive

cpitclaudel:
This prevents the transaction highlight from interfering with the region
highlight, as for example in the following case:

```
<1>2017-01-01 * Test
    Expenses:A  $15
    Income:B

2017-01-01 * Test
    Expenses:A  $15
    Income:B
<0>
```

(Place point on <0>, then press C-SPC and move the point to <1>: note how
the transaction overlay takes precedence over the region overlay, which is undesirable)

### DIFF
--- a/ledger-xact.el
+++ b/ledger-xact.el
@@ -56,8 +56,9 @@
       (let ((b (car exts))
             (e (cadr exts))
             (p (point)))
-        (if (and (> (- e b) 1)       ; not an empty line
-                 (<= p e) (>= p b))  ; point is within the boundaries
+        (if (and (> (- e b) 1)            ; not an empty line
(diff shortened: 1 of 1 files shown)


purcell:
I'm not entirely sure if this is the right general fix - the problem is more that the faces on your system are apparently such that you can't distinguish the region face from the highlight face. That's not the case on my system, for example, so it may just be a quirk of your theme. This *may* still be a useful fix... just not certain.

cpitclaudel:
> the problem is more that the faces on your system are apparently such that you can't distinguish the region face from the highlight face

No, that's not it; the issue is with overlay priorities:

rec3

purcell:
Cool, thanks for the clarification. Merging this, then!

PR #73, opened 2018-01-22
Improve highlighting in reports

cpitclaudel:
* Add a button face and enable clicking on register report lines
* Use ledger's own highlighting in report buffers

Supersedes GH-68? CC @jabranham

### DIFF
--- a/doc/ledger-mode.texi
+++ b/doc/ledger-mode.texi
@@ -1026,6 +1026,9 @@ functions.
 Whether reports should be displayed using the same colors as when calling ledger
 on the command line.
 
+@item ledger-report-auto-width
+Whether reports should fill the whole width of the @file{*Report*} window
+
 @end ftable
 
 @node Ledger Faces Customization Group, Ledger Post Customization Group, Ledger Report Customization Group, Customization Variables
--- a/doc/ledger-mode.texi
+++ b/doc/ledger-mode.texi
@@ -1022,6 +1022,10 @@ Definition of reports to run.
 An alist mapping ledger report format specifiers to implementing
 functions.
 
+@item ledger-report-use-native-highlighting
+Whether reports should be displayed using the same colors as when calling ledger
+on the command line.
+
 @end ftable
 
 @node Ledger Faces Customization Group, Ledger Post Customization Group, Ledger Report Customization Group, Customization Variables
@@ -1080,7 +1084,7 @@ Default face for pending @samp{!} transactions in the @file{*Reconcile*}
 buffer.
 
 @item ledger-font-report-clickable-face
-Face for transactions in the @file{*Report*} buffer.
+Face applied to clickable entries in the @file{*Report*} buffer.
 
 @end ftable
 
--- a/ledger-fonts.el
+++ b/ledger-fonts.el
@@ -261,8 +261,8 @@
   :group 'ledger-faces)
 
 (defface ledger-font-report-clickable-face
-  `((t :inherit ledger-font-default-face))
-  "Face for transactions in the report buffer."
+  `((t))
+  "Face applied to clickable entries in the report window"
   :group 'ledger-faces)
 
 (defface ledger-font-code-face
--- a/ledger-report.el
+++ b/ledger-report.el
@@ -149,7 +149,6 @@ in the `header-line'."
       'ledger-report-kill)
     (define-key map [(control ?c) (control ?l) (control ?e)]
       'ledger-report-edit)
-    (define-key map [return] 'ledger-report-visit-source)
     map)
   "Keymap for `ledger-report-mode'.")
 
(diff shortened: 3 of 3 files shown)


jabranham:
Nice! This is definitely better than #68 since it also fixes #20, something @akater has been working on, I believe. Can you also update the description of ledger-font-report-clickable-face in doc/ledger-mode.texi line 1082?

cpitclaudel:
Done, thanks for the pointer :)

cpitclaudel:
Btw, what do you think of the following additional patch?

```diff
diff --git a/ledger-report.el b/ledger-report.el
index 7efde22..2ceb69e 100644
--- a/ledger-report.el
+++ b/ledger-report.el
@@ -93,6 +93,11 @@ ledger-report-use-native-highlighting
   :type 'boolean
   :group 'ledger-report)
 
+(defcustom ledger-report-auto-width t
+  "When non-nil, tell ledger about the width of the report window."
+  :type 'boolean
+  :group 'ledger-report)
+
 (defcustom ledger-report-use-header-line nil
   "When non-nil, indicate the report name and command in the `header-line'
 instead of in the buffer."
@@ -379,6 +384,8 @@ ledger-do-report
       (setq cmd (concat cmd " --prepend-format='%(filename):%(beg_line):'")))
     (when ledger-report-use-native-highlighting
       (setq cmd (concat cmd " --color --force-color")))
+    (when ledger-report-auto-width
+      (setq cmd (concat cmd (format " --columns %d" (window-width)))))
     (shell-command cmd t nil)
     (when ledger-report-use-native-highlighting
       (ansi-color-apply-on-region data-pos (point-max)))
```

akater:
Looks like this patch merely appends formatting options to the tail of command string. This won't work well for commands which contain multiple calls to ledger which is why I used regex in my approach; see example of multi-call command here https://github.com/ledger/ledger-mode/pull/61#issuecomment-356431289 I wouldn't call this a corner case, as I myself use a custom report like that one daily, more or less. Certainly useful for multicurrency journals.

cpitclaudel:
> Looks like this patch merely appends formatting options to the tail of command string. This won't work well for commands which contain multiple calls to ledger which is why I used regex in my approach; see example of multi-call command here #61 (comment) I wouldn't call this a corner case, as I myself use a custom report like that one daily, more or less. Certainly useful for multicurrency journals.

Great point. What do you think of the fix I just pushed?  The idea is to include these arguments into the expansion of `%(binary)`

akater:
Clément Pit-Claudel <notifications@github.com> writes:

> Great point. What do you think of the fix I just pushed?  The idea is to include these arguments into the expansion of `%(binary)`
This is of course much cleaner than what I suggested. I can't review all
of it atm but as soon as it does well for your everyday needs, I
will cancel my pull request, proposing to merge yours instead.

One minor note: your patch deals both with highlighting and report
width. I tested my report width patch in both tty Emacs and X (GUI)
Emacs, and found that it required a little adjustment: see line 363 here
https://github.com/ledger/ledger-mode/compare/master...akater:width-auto-adjustment#diff-281a2bc801a5ecafb82d5fa789415bb2R363

Not sure if this is reproducible but if you care about tty Emacs, you
might want to take a look.

cpitclaudel:
> This is of course much cleaner than what I suggested. I can't review all
of it atm but as soon as it does well for your everyday needs, I
will cancel my pull request, proposing to merge yours instead.

Thanks! Hopefully we can get an extra review before merging :)

>  One minor note: your patch deals both with highlighting and report width. I tested my report width patch in both tty Emacs and X (GUI) Emacs, and found that it required a little adjustment: see line 363 here master...akater:width-auto-adjustment#diff-281a2bc801a5ecafb82d5fa789415bb2R363

I actually left the report width part out.  I originally didn't see your PR, but I took it out once I did :) I'll reintegrate it, but I'll proper credits.

> found that it required a little adjustment: see line 363 here master...akater:width-auto-adjustment#diff-281a2bc801a5ecafb82d5fa789415bb2R363

I left a comment on the PR.  I think using window-width for everything is good, and it should need a -1 everywhere.

cpitclaudel:
> I'll reintegrate it, but I'll proper credits.

Done.  I had to push an extra commit on top of that to make sure that refreshing the report window with `g` would recompute the window width.

akater:
Great. Maybe you need to `(require 'ansi-color)` somewhere.

cpitclaudel:
> 
> Great. Maybe you need to (require 'ansi-color) somewhere.

Yes, excellent point.

purcell:
Looks good to me - merged!

purcell:
May I suggest to @jwiegley / @enderw88 that we add @cpitclaudel as a collaborator on this repo please?

jwiegley:
I've invited you and he to become part of the Ledger Hackers group, which will give you full admin rights to all three related repositories.

purcell:
Thanks @jwiegley - for a while now I've been following your example of handing out the commit bit liberally.

akater:
@cpitclaudel Just one minor thing I forgot to mention. The expression `(- window-width 1)` is mentioned twice in the current version of legder-report.el. In both cases it is computed for essentially the same purpose. I'd say it's a clear violation of D.R.Y. to leave these two as is; better to define a function.

PR #74, opened 2018-01-22
Don't force completion in ledger-completing-read-with-default

cpitclaudel:
Forcing completion makes it impossible to use regexps or partial names (at least
with IDO).


purcell:
Makes sense and looks good - thanks!

PR #75, opened 2018-01-22
All defcustoms should specify :type

jabranham:
This becomes a byte compiler warning in Emacs 26


purcell:
Thanks. I think that first one can be `'file`, right?

jabranham:
Fixed, thanks

purcell:
Thanks!

PR #76, opened 2018-01-24
Fix Travis builds

cpitclaudel:
This required the following changes:

* Dropping the OSX builds: the Travis OSX builders are saturated.

* Reverting to the version of Boost contained in the official Ubuntu repos; that's one version behind the one in the PPA, but that doesn't seem to be a problem.  The PPA wasn't updated to provide trusty builds.

* Disabling warnings-as-errors on Emacs 24.3, since that version of Emacs produces many false positives that get turned into hard errors.

* Fixing ~25 test failures: two in context-tests (a new feature was implemented but the test had not been updated, and string-greater-p was used without providing a shim for Emacs 24.3, which doesn't have that function), and the rest in fontify-tests (most of these were of the following form: the test expected the newline following a string to be fontified, but the current version of the text doesn't fontify the newline — this happens e.g. on directive and account lines)

I also took the opportunity to speed up the builds by caching the ledger binaries; this saves roughly 20 minutes per build, so builds now take 2-4 minutes instead of 22-24 minutes (see e.g. https://travis-ci.org/cpitclaudel/ledger-mode/builds/332623191)

### DIFF
--- a/ledger-complete.el
+++ b/ledger-complete.el
@@ -25,6 +25,13 @@
 (require 'pcomplete)
 (require 'cl-lib)
 
+(unless (fboundp 'string-greaterp) ;; Emacs 24.3 compatibility
+  (defun string-greaterp (string1 string2)
+    "Return non-nil if STRING1 is greater than STRING2 in lexicographic order.
+Case is significant.
+Symbols are also allowed; their print names are used instead."
+    (string-lessp string2 string1)))
+
 ;; In-place completion support
 
 ;;; Code:
--- a/test/Makefile
+++ b/test/Makefile
@@ -13,7 +13,7 @@ CHECKDOC_BATCH_EL := ../tools/checkdoc-batch.el
 all: compile test
 
 %.elc: %.el
-	$(EMACS_BATCH) --eval "(progn (setq byte-compile-error-on-warn t) (batch-byte-compile))" $<
+	$(EMACS_BATCH) --eval "(progn (setq byte-compile-error-on-warn (>= emacs-major-version 25)) (batch-byte-compile))" $<
 
 .PHONY: test
 test: test-interactive test-batch
--- a/tools/travis-install-ledger.sh
+++ b/tools/travis-install-ledger.sh
@@ -6,20 +6,25 @@ set -o xtrace
 BRANCH=$1
 
 if [ "${TRAVIS_OS_NAME}" = "linux" ]; then
-    sudo add-apt-repository -y ppa:ubuntu-toolchain-r/test
-    sudo apt-get update -qq
-    sudo apt-get install -qq gcc-4.8 g++-4.8
-    sudo update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-4.8 90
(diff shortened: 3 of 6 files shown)


ledger-complete.el
```diff
@@ -25,6 +25,13 @@
 (require 'pcomplete)
 (require 'cl-lib)
 
+(unless (fboundp 'string-greaterp) ;; Emacs 24.3 compatibility
```
purcell: Note that the [elisp coding conventions](https://www.gnu.org/software/emacs/manual/html_node/elisp/Coding-Conventions.html) say not to shim functions for backwards-compatibility like this. It would probably be preferable to use `string<` throughout, swapping args of course.

test/Makefile
```diff
@@ -13,7 +13,7 @@ CHECKDOC_BATCH_EL := ../tools/checkdoc-batch.el
 all: compile test
 
 %.elc: %.el
-	$(EMACS_BATCH) --eval "(progn (setq byte-compile-error-on-warn t) (batch-byte-compile))" $<
+	$(EMACS_BATCH) --eval "(progn (setq byte-compile-error-on-warn (>= emacs-major-version 25)) (batch-byte-compile))" $<
```
purcell: This adjustment was for Emacs 24.3, right, but what about 24.4 / 24.5?

tools/travis-install-ledger.sh
```diff
@@ -1,27 +1,45 @@
...
+        fi
+
+        cd "ledger-$BRANCH"
+
+        git fetch origin
+        git reset --hard "origin/$BRANCH"
+
+        REV="$(git rev-parse "$BRANCH")"
+
+        if [ "$(cat "$LAST_BUILD_FILE" 2> /dev/null)" = "$REV" ]; then
```
purcell: Would `make` not handle this case correctly? It's not only `$REV` which determines the results of the build, because you might switch compiler or boost version etc., so this check seems a little too aggressive IMO.

ledger-complete.el
cpitclaudel: OK, will fix :)

test/Makefile
cpitclaudel: I just assumed that being warning-free on the latest release was enough — especially since we don't currently test on 24.4 or 24.5.  I can make the test more precise using `version<` if you want

tools/travis-install-ledger.sh
cpitclaudel: I think it would, but it'd require installing more stuff (adding an extra PPA and installing GCC from there), which takes extra time for nothing in most cases.

> you might switch compiler or boost version etc.

Will `make` detect that?  I'd opt for speed over extra caution for compiling ledger, since ledger build issues aren't this repo's responsibility :) And Travis makes it easy to clear the cache to do a full rebuild if desired.

WDYT?

test/Makefile
purcell: Nah, I guess it's fine. Just wanted to check.

tools/travis-install-ledger.sh
purcell: Yep, probably agree. Make *should* detect that case, but I'm unsure, and this is probably the correct trade-off.

PR #77, opened 2018-01-25
Don't shim string-greaterp

cpitclaudel:
This didn't make it into the previous PR in time :)


purcell:
All good, sorry for rushing. :-)

PR #78, opened 2018-01-26
Lexical bindings

jabranham:
Prefer `lexical-binding t`

This started out as an attempt to fix a few compiler warnings, which I'll submit later. The compiler couldn't find `lexical-let*` as a function so I decided to remove it rather than require cl at compile


purcell:
Excellent, thank you!

PR #81, opened 2018-01-30
let-bind buffer-read-only

jabranham:
setq'ing it means the buffer wasn't read-only before this e.g. after pressing `e` to edit a report


cpitclaudel:
I think the usual way to do this is to let-bind inhibit-read-only; wdyt?

jabranham:
I pushed a new commit doing that. I used buffer-read-only before since it was there already and is used a few other places.

cpitclaudel:
Super, thanks!

PR #82, opened 2018-01-30
Improve ledger-report-reverse-lines

jabranham:
- let-bind inhibit-read-only so that the report is put back in
  read-only-mode
- don't call forward-paragraph if the report is in the header rather
  than in the buffer

### DIFF
--- a/ledger-report.el
+++ b/ledger-report.el
@@ -134,12 +134,14 @@ in the `header-line'."
   (setq ledger-report-is-reversed (not ledger-report-is-reversed)))
 
 (defun ledger-report-reverse-lines ()
-  (goto-char (point-min))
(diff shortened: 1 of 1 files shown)


cpitclaudel:
Thanks; that's a great change.  I wish we were doing something more principled (like having a marker at the beginning of the report area, or a text property covering it) - maybe we'll address that in a future refactoring.

Maybe we could add a test too, now that we have CI working again.

jabranham:
That all sounds good.

I also noticed that the native highlighting doesn't work once you reverse the report.

cpitclaudel:
I think that's worth a bug report, unless you have a quick fix in the works :)

Issue #83, opened 2018-01-30
Native highlighting in reports does not work after reversing lines


jabranham:
And the fix isn't as simple as reapplying `ansi-color-apply-on-region`.

PR #84, opened 2018-01-31
Fix GH-83 by applying highlights as text properties, not overlays

cpitclaudel:
CC @jabranham.
The crucial bit is using ansi-color-apply instead of ansi-color-apply-on-region


jabranham:
Neat, thanks. I haven't worked much with `ansi-color.el`.

PR #86, opened 2018-02-05
Fix typo in ledger-mode.texi

NateEag:
Just replacing 'it' with 'its'.

Noticed this while reading the docs and thought I'd submit a quick patch.


cpitclaudel:
Thanks!

PR #87, opened 2018-02-06
Add support for flymake in Emacs 26+

jabranham:
Flymake, Emacs's built-in on-the-fly syntax checker, got a major
overhaul in Emacs 26. This means we can get built-in syntax checking
without requiring users to install both flycheck.el and
flycheck-ledger.el. The checking is done via `ledger balance`. Things
to keep in mind that as of now (the initial implementation):

- It will honor settings from users ~/.ledgerrc files (I'm not sure if
  this is a good thing or a bad thing)

- ~It will error if Emacs can't find ledger (i.e. if (executable-find
  "ledger") fails)~ Now uses `ledger-binary-path`

- ~There is no way yet to customize calls (using --pedantic, for
  example), though this would be easy to implement in the future~ Supports `--pedantic` and `--explicit`

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,8 +32,10 @@ quick demo as well as more extensive documentation.
 Related packages
 ================
 
-In-buffer checking of formatting and balancing of transactions is
-available using [`flycheck-ledger`][flycheck-ledger].
+In-buffer checking of formatting and balancing of transactions is available
+built-in for Emacs version 26 and later using `flymake-mode`. For flycheck users
+(and users of Emacses earlier and 26), [`flycheck-ledger`][flycheck-ledger] is
+available.
 
 
 
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@ Related packages
 
 In-buffer checking of formatting and balancing of transactions is available
 built-in for Emacs version 26 and later using `flymake-mode`. For flycheck users
-(and users of Emacses earlier and 26), [`flycheck-ledger`][flycheck-ledger] is
+(and users of Emacs 25 and earlier), [`flycheck-ledger`][flycheck-ledger] is
 available.
 
 
--- a/ledger-flymake.el
+++ b/ledger-flymake.el
@@ -27,6 +27,9 @@
 ;;; Code:
 (require 'cl-lib)
 (require 'ledger-exec)                  ; for `ledger-binary-path'
+(declare-function flymake-diag-region "flymake")
+(declare-function flymake-make-diagnostic "flymake")
+(declare-function flymake-log "flymake")
 
 (defvar-local ledger--flymake-proc nil)
 
--- a/ledger-flymake.el
+++ b/ledger-flymake.el
@@ -21,7 +21,7 @@
 
 ;;; Commentary:
 ;; Flymake is the built-in Emacs package to support on-the-fly syntax checking.
-;; This file adds support for flymake to `ledger-mode'. Enable it by calling
+;; This file adds support for flymake to `ledger-mode'.  Enable it by calling
 ;; `flymake-mode' from a file-visiting ledger buffer.
 
 ;;; Code:
@@ -90,7 +90,7 @@ Flymake calls this with REPORT-FN as needed."
                       ;; of objects, and call `report-fn'.
                       (cl-loop
                        while (search-forward-regexp
-                              ;; This regex needs to match the whole error. We
+                              ;; This regex needs to match the whole error.  We
                               ;; also need a capture group for the error message
                               ;; (that's group 1 here) and the line number
                               ;; (group 2).
--- a/ledger-flymake.el
+++ b/ledger-flymake.el
@@ -27,6 +27,7 @@
 ;;; Code:
 
 (defvar-local ledger--flymake-proc nil)
+(defvar ledger-binary-path)
 
 ;; Based on the example from Flymake's info:
 (defun ledger-flymake (report-fn &rest _args)
@@ -100,7 +101,7 @@ Flymake calls this with REPORT-FN as needed."
               (kill-buffer (process-buffer proc))))))))))
 
 (defun ledger-enable-flymake ()
-  "Enable `flymake-mode' in ledger-mode buffers.
+  "Enable function `flymake-mode' in `ledger-mode' buffers.
 
 Don't enable flymake if flycheck is on and flycheck-ledger is
 available."
--- a/ledger-flymake.el
+++ b/ledger-flymake.el
@@ -25,8 +25,7 @@
 ;; `flymake-mode' from a file-visiting ledger buffer.
 
 ;;; Code:
-(eval-when-compile
-  (require 'cl-lib))
+(require 'cl-lib)
 (require 'ledger-exec)                  ; for `ledger-binary-path'
 
 (defvar-local ledger--flymake-proc nil)
--- a/ledger-flymake.el
+++ b/ledger-flymake.el
@@ -99,13 +99,22 @@ Flymake calls this with REPORT-FN as needed."
               ;; check's output.
               (kill-buffer (process-buffer proc))))))))))
 
+(defun ledger-enable-flymake ()
+  "Enable `flymake-mode' in ledger-mode buffers.
+
+Don't enable flymake if flycheck is on and flycheck-ledger is
+available."
+  (unless (and (featurep 'flycheck-ledger)
+               (bound-and-true-p flycheck-mode)))
+  (flymake-mode))
+
 ;; Automatically enable flymake in ledger-mode buffers when using Emacs 26+
 (when (<= 26 emacs-major-version)
   ;; Add `ledger-flymake' to `flymake-diagnostic-functions' so that flymake can
   ;; work in ledger-mode:
   (add-hook 'flymake-diagnostic-functions 'ledger-flymake nil t)
   ;; Enable it!
-  (add-hook 'ledger-mode-hook #'flymake-mode))
+  (add-hook 'ledger-mode-hook #'ledger-enable-flymake))
 
 (provide 'ledger-flymake)
 
--- a/ledger-flymake.el
+++ b/ledger-flymake.el
@@ -70,8 +70,8 @@ Flymake calls this with REPORT-FN as needed."
         :name "ledger-flymake" :noquery t :connection-type 'pipe
         :buffer (generate-new-buffer " *ledger-flymake*")
         :command `(,ledger-binary-path "-f" ,file
-                                       (when ledger-flymake-be-pedantic "--pedantic")
-                                       (when ledger-flymake-be-explicit "--explicit")
+                                       ,(when ledger-flymake-be-pedantic "--pedantic")
+                                       ,(when ledger-flymake-be-explicit "--explicit")
                                        "balance")
         :sentinel
         (lambda (proc _event)
--- a/ledger-flymake.el
+++ b/ledger-flymake.el
@@ -29,6 +29,20 @@
 (defvar-local ledger--flymake-proc nil)
 (defvar ledger-binary-path)
 
+(defcustom ledger-flymake-be-pedantic nil
+  "If non-nil, pass the --pedantic flag for ledger to the flymake backend.
+If --pedantic is in your ledgerrc file, then --pedantic gets
+passed regardless of the value."
+  :type 'boolean
+  :group 'ledger)
+
+(defcustom ledger-flymake-be-explicit nil
+  "If non-nil, pass the --explicit flag for ledger to the flymake backend.
+If --explicit is in your ledgerrc file, then --explicit gets
+passed regardless of the value."
+  :type 'boolean
+  :group 'ledger)
+
 ;; Based on the example from Flymake's info:
 (defun ledger-flymake (report-fn &rest _args)
   "A Flymake backend for `ledger-mode'.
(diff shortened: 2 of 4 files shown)


jabranham:
Pinging @purcell as the author of the great flycheck-ledger package.

ledger-flymake.el
```diff
@@ -0,0 +1,114 @@
...
+              ;; check's output.
+              (kill-buffer (process-buffer proc))))))))))
+
+;; Automatically enable flymake in ledger-mode buffers when using Emacs 26+
+(when (version<= "26" emacs-version)
+  ;; Add `ledger-flymake' to `flymake-diagnostic-functions' so that flymake can
+  ;; work in ledger-mode:
+  (add-hook 'flymake-diagnostic-functions 'ledger-flymake nil t)
+  ;; Enable it!
+  (add-hook 'ledger-mode-hook #'flymake-mode))
```
cpitclaudel: I'd be careful about this: won't users of flycheck get two competing backends?

ledger-flymake.el
```diff
@@ -0,0 +1,114 @@
...
+                       into diags
+                       finally (funcall report-fn diags)))
+                  (flymake-log :warning "Canceling obsolete check %s"
+                               proc))
+              ;; Cleanup the temporary buffer used to hold the
+              ;; check's output.
+              (kill-buffer (process-buffer proc))))))))))
+
+;; Automatically enable flymake in ledger-mode buffers when using Emacs 26+
+(when (version<= "26" emacs-version)
```
cpitclaudel: You can use emacs-major-version here, I think

ledger-flymake.el
jabranham: They don't "compete", at least not negatively. Sure, both might be turned on but that won't mess anything up. The call to ledger balance is pretty minimal, so it shouldn't impact performance. I think the worst that could happen is that the two faces get combined into some horrible monstrosity. 

But it's easy to turn one off.

I wanted to enable it by default since it seems like something that would benefit most users. If yall want though I can leave that out.

ledger-flymake.el
jabranham: I get an error if I do that: `(error "Version must be a string")`

jabranham: Is there a reason to prefer that to what I have already, BTW?

cpitclaudel: I meant `(<= 26 emacs-major-version)`; this saves a call to version<=, that's all :)

ledger-flymake.el
cpitclaudel: I'm fine with enabling it by default; I was suggesting something like (unless (bound-and-true-p flycheck-mode) …)

ledger-flymake.el
jabranham: Done! 👍

ledger-flymake.el
jabranham: But if we do that, people who have called `flycheck-global-mode` but who haven't installed `flycheck-ledger` won't get this new flymake stuff, right?

How about if we do `(unless (featurep 'flycheck-ledger) ...)`?

cpitclaudel: Looks good.  I'm nto sure what's the standard etiquette about enabling flymake-mode, but I like that we at least won't run both at the same time.

cpitclaudel:
LGTM. Should me merge flycheck-ledger into flycheck-mode as well?

jabranham:
I just pushed a final commit that makes `checkdoc` and the byte compiler happy. Should be ready, I think. 

Thanks!

README.md
```diff
@@ -32,8 +32,10 @@ quick demo as well as more extensive documentation.
 Related packages
 ================
 
-In-buffer checking of formatting and balancing of transactions is
-available using [`flycheck-ledger`][flycheck-ledger].
+In-buffer checking of formatting and balancing of transactions is available
+built-in for Emacs version 26 and later using `flymake-mode`. For flycheck users
+(and users of Emacses earlier and 26), [`flycheck-ledger`][flycheck-ledger] is
```
cpitclaudel: "Emacsen", I think (?)

cpitclaudel:
Neat. Let's wait to hear from @purcell ; I don't know anything about the new Flymake, so it's hard for me to review this.

README.md
jabranham: I thought Emacsen referred to GNU Emacs plus all the various forks and other versions (xemacs, for example). But either way the wording is horrible right now. I'll push a commit in a minute that fixes it.

jabranham:
Also, the combination of this and `flycheck-ledger` probably means that `ledger-check.el` (which is mostly broken for me anyway) can be made obsolete/removed.

purcell:
> Should me merge flycheck-ledger into flycheck-mode as well?

No: then `ledger-mode` would have to depend on `flycheck` for all its files to compile cleanly.

ledger-flymake.el
```diff
@@ -0,0 +1,139 @@
...
+              ;; check's output.
+              (kill-buffer (process-buffer proc))))))))))
+
+(defun ledger-enable-flymake ()
+  "Enable function `flymake-mode' in `ledger-mode' buffers.
+
+Don't enable flymake if flycheck is on and flycheck-ledger is
+available."
+  (unless (and (featurep 'flycheck-ledger)
+               (bound-and-true-p flycheck-mode)))
```
purcell: I don't think this check is really the way to go: you're looking for competing packages in order to make the default behaviour "on", which doesn't smell right. You're even enabling `flymake-mode`, even if the user doesn't like getting in-buffer feedback.

Better, then, would be to provide this function *without* this check, and add an `;;;###autoload` cookie for it, then you'd remove the `(when ...)` form below, and note in the documentation instead that `flymake` users can add the function to `ledger-mode-hook` (e.g. even by using `M-x customize-variable`).

ledger-mode.el
```diff
@@ -53,6 +53,7 @@
 (require 'ledger-xact)
 (require 'ledger-schedule)
 (require 'ledger-check)
+(require 'ledger-flymake)
```
purcell: And then you'd drop this, I think. This feature needn't ever be loaded by `flycheck` users.

ledger-flymake.el
```diff
@@ -0,0 +1,139 @@
...
+                ;; Only proceed if `proc' is the same as
+                ;; `ledger--flymake-proc', which indicates that
+                ;; `proc' is not an obsolete process.
+                (if (with-current-buffer source (eq proc ledger--flymake-proc))
+                    (with-current-buffer (process-buffer proc)
+                      (goto-char (point-min))
+                      ;; Parse the output buffer for diagnostic's
+                      ;; messages and locations, collect them in a list
+                      ;; of objects, and call `report-fn'.
+                      (cl-loop
```
purcell: You should `(require 'cl-lib)` for this.

ledger-flymake.el
```diff
@@ -0,0 +1,139 @@
...
+;; Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston,
+;; MA 02110-1301 USA.
+
+;;; Commentary:
+;; Flymake is the built-in Emacs package to support on-the-fly syntax checking.
+;; This file adds support for flymake to `ledger-mode'.  Enable it by calling
+;; `flymake-mode' from a file-visiting ledger buffer.
+
+;;; Code:
+
```
purcell: Wouldn't you need a `(require 'flymake)` form?

ledger-flymake.el
```diff
@@ -0,0 +1,139 @@
...
+(defvar-local ledger--flymake-proc nil)
+(defvar ledger-binary-path)
```
purcell: You can probably just `require` the feature that defines this variable, right?

ledger-flymake.el
```diff
@@ -0,0 +1,139 @@
...
+                       into diags
+                       finally (funcall report-fn diags)))
+                  (flymake-log :warning "Canceling obsolete check %s"
+                               proc))
+              ;; Cleanup the temporary buffer used to hold the
+              ;; check's output.
+              (kill-buffer (process-buffer proc))))))))))
+
+(defun ledger-enable-flymake ()
+  "Enable function `flymake-mode' in `ledger-mode' buffers.
```
purcell: Is the "function" word here just to appease `checkdoc`? :-)

ledger-mode.el
purcell: Also, this way you could check for a compatible `flymake` version at the top of `ledger-flymake`.

jabranham:
> Big +1 for this in principle, if it compiles cleanly in all the supported emacsen. I noted a few things along the way.

Thanks! I replied to everything, I think, let me know if I missed something.

 I'm not sure the best way to see if it compiles cleanly in all supported emacs versions. It looks good on my end using the Emacs 26 pretest. Do you have access to Emacs 24.3 or 25.3?

If compiling cleanly on all supported Emacs versions is something that all PRs should do, perhaps that could be added as a test on Travis?

ledger-flymake.el
purcell: > I can wrap it in (eval-when-compile ...) though since it's a macro, right? That's what I've done, let me know if it's not right.

No, you wouldn't normally do that with `cl-lib`: just `require` it.

ledger-flymake.el
purcell: > But if you want, I can turn if off and make these changes. Just let me know.

Yes please.

purcell:
> If compiling cleanly on all supported Emacs versions is something that all PRs should do, perhaps that could be added as a test on Travis?

I believe it is already there: https://github.com/ledger/ledger-mode/blob/master/test/Makefile#L16

ledger-flymake.el
cpitclaudel: But `cl-loop` is autoloaded from `cl-macs`, right?  Does `cl-lib` help with it?
IOW, wouldn't it be enough to `(eval-when-compile (require 'cl-macs))`?

purcell: > But cl-loop is autoloaded from cl-macs, right? Does cl-lib help with it?

It's part of `cl-lib`, so generally if you functions from `cl-lib`, you should explicitly `require` it.

> IOW, wouldn't it be enough to (eval-when-compile (require 'cl-macs))?

Nooooo - `cl-macs` should never be required explicitly. If `cl-lib` is installed as a separate package, it's in a single file, and `cl-macs.el` will not exist.

cpitclaudel: > It's part of cl-lib, so generally if you functions from cl-lib, you should explicitly require it.

I think. I guess the part about the separate package version is crucial for this part too, right? Because it seems that in the bundled version, loading cl-lib doesn't load cl-loop (that is, cl-macs requires cl-lib, but not the other way around).

> Nooooo - cl-macs should never be required explicitly. If cl-lib is installed as a separate package, it's in a single file, and cl-macs.el will not exist.

… wooops.  And there I was, thinking I learned a cool trick from reading the source code of thunk.el...  Well, I learned something, thanks for taking the time to explain!

purcell: > I think. I guess the part about the separate package version is crucial for this part too, right? Because it seems that in the bundled version, loading cl-lib doesn't load cl-loop (that is, cl-macs requires cl-lib, but not the other way around).

Yeah, the way to think about it is that after loading either the built-in or ELPA versions of `cl-lib`, the same set of functions/macros will be available. That some of them are available *without* requiring that feature is not something to rely on IMO.

jabranham:
OK, I think I've covered everything. Let me know if I missed something. Thanks for the comments!

ledger-flymake.el
purcell: You can't really know from looking at code built into Emacs itself because that can always assume that the code around it is at a certain version - e.g. python.el in Emacs 26 can be written to assume that flymake is the one from Emacs 26.

jabranham: I added it in the meantime :-)

purcell: Looks like these aren't available in the Emacs snapshot version on Travis, so byte compilation fails: did the build pass previously? If so, let's just forget this last commit. Keen to get this merged now -- it's looking great.

jabranham: I think it's failing because the branch was created before the byte compilation got fixed in #91 and the the related commit. If you want, I can rebase with those changes, then the CI should pass. Lemme know.

purcell:
> I think it's failing because the branch was created before the byte compilation got fixed in #91 <https://github.com/ledger/ledger-mode/pull/91> and the the related commit. If you want, I can rebase with those changes, then the CI should pass. Lemme know.
> 
> 


Yes please, I think that’d be a good plan.

jabranham:
Rebased onto `master`

jabranham:
It looks to me like the `emacs-snapshot` build is failing because the snapshot is too old. The copyright year in `emacs --version` is 2016, but the flymake rewrite got merged in 2017 sometime.

jabranham:
I'm not super familiar with Ubuntu or Travis, but it looks like to me that the [ubuntu-elisp ppa](https://launchpad.net/~ubuntu-elisp/+archive/ubuntu/ppa)'s latest emacs-snapshot build for Ubuntu 14.04 (which is what travis runs) was built in 2016. I've pushed a commit that allows this build to fail.

It actually looks like it was supposed to be allowed to fail anyway.

purcell:
> It actually looks like it was supposed to be allowed to fail anyway.

Yep, good catch. Merged now - thanks!!

Re. the Travis issues, we could maybe switch to using EVM, which I've used in Travis for a bunch of other elisp projects, but it seems somewhat deprecated now and had some issues on MacOS IIRC.

PR #88, opened 2018-02-08
Fixes for the byte compiler

jabranham:
Most of these are new warnings from the switch to lexical-binding in 6c2e8bdf

### DIFF
--- a/ledger-check.el
+++ b/ledger-check.el
@@ -119,8 +119,7 @@ commands for navigating the buffer to the errors found, etc."
      (when (and (buffer-modified-p)
                 (y-or-n-p "Buffer modified, save it? "))
        (save-buffer))))
-  (let ((buf (find-file-noselect (ledger-master-file)))
-        (cbuf (get-buffer ledger-check-buffer-name))
+  (let ((cbuf (get-buffer ledger-check-buffer-name))
         (wcfg (current-window-configuration)))
     (if cbuf
         (kill-buffer cbuf))
--- a/ledger-xact.el
+++ b/ledger-xact.el
@@ -86,7 +86,7 @@ MOMENT is an encoded date"
     (catch 'found
       (ledger-xact-iterate-transactions
        (function
-        (lambda (start date mark desc)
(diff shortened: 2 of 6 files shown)


ledger-xact.el
```diff
@@ -122,11 +121,12 @@ MOMENT is an encoded date"
...
 
 (defun ledger-copy-transaction-at-point (date)
-  "Ask for a new DATE and copy the transaction under point to that date.  Leave point on the first amount."
+  "Ask for a new DATE and copy the transaction under point to that date.
+Leave point on the first amount."
   (interactive  (list
                  (ledger-read-date "Copy to date: ")))
   (let* ((here (point))
-         (extents (ledger-navigate-find-xact-extents (point)))
+         (extents (ledger-navigate-find-xact-extents here))
```
cpitclaudel: Is `here` used elsewhere? Otherwise, I think we can remove it.

ledger-check.el
```diff
@@ -119,8 +119,7 @@ commands for navigating the buffer to the errors found, etc."
      (when (and (buffer-modified-p)
                 (y-or-n-p "Buffer modified, save it? "))
        (save-buffer))))
-  (let ((buf (find-file-noselect (ledger-master-file)))
-        (cbuf (get-buffer ledger-check-buffer-name))
+  (let ((cbuf (get-buffer ledger-check-buffer-name))
```
cpitclaudel: That code (the existing one, not yours) looks very suspicious.  Does it really run the check in an empty buffer?  Why did it open the file for nothing?

ledger-xact.el
jabranham: Not that I could see. I left it there just in case, since it doesn't really cost anything.

ledger-check.el
jabranham: I'm not 100% certain. 

Like I said in #87 ledger-check looks broken on my end. There's not much documentation for it either, so I'm not really sure what it's *supposed* to do/look like.

PR #89, opened 2018-02-09
Ledger master buffer local

jabranham:
This PR makes ledger-master-file a buffer-local variable, which is what the documentation of ledger-master-file (the function) says it is.

Also some bonus documentation changes/improvements mostly to satisfy checkdoc.

### DIFF
--- a/ledger-report.el
+++ b/ledger-report.el
@@ -303,7 +303,9 @@ used to generate the buffer, navigating the buffer, etc."
 
 ;; General helper functions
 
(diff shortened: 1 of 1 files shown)


cpitclaudel:
Thanks! Interestingly, we never set that variable ourselves.

purcell:
> Interestingly, we never set that variable ourselves.

Yep, I think this should be a `defcustom`, actually.

PR #91, opened 2018-02-09
Make travis error on byte compiler warnings

jabranham:
Adds a test that fails if ledger-mode doesn't byte-compile without warnings/errors. 

It looks like the ubuntu-elisp ppa updated to Emacs 26 at some point. I can't figure out how to add back in a test for Emacs 25.

The tests will of course fail right now since #88 hasn't been merged :-)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,11 +15,11 @@ env:
   matrix:
     - EMACS_VERSION=emacs24 LEDGER_BRANCH=master
     # - EMACS_VERSION=emacs24 LEDGER_BRANCH=apt-get # apt-get's ledger is too old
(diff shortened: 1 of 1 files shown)


cpitclaudel:
Neat. But we already have a line in the makefile for that, normally: https://github.com/ledger/ledger-mode/blob/7a3d71d163a14203d39774b8e7245a6a2a909370/test/Makefile#L16

Is it broken?

jabranham:
Since it's run from the "test" directory, it's byte-compiling the test files, not the ledger-mode files.

cpitclaudel:
Ah, of course 😖

purcell:
Some fixes for these in #95.

purcell:
NICE!

Issue #94, opened 2018-02-11
Build system does not ignore site- and user-specific init files


scfc:
The build system currently does not run Emacs with `--no-site-file` and `--no-init-file`, potentially being influenced by local customizations (for better or worse).

For `ledger-mode` as part of `ledger`, I came up with a patch against `lisp/CMakeLists.txt` to call Emacs as `EMACSLOADPATH= ${EMACS_EXECUTABLE} […] -batch --no-init-file --no-site-file -f batch-byte-compile`.  This still applies (cf. https://github.com/scfc/ledger-mode/tree/ignore-init-files-when-byte-compiling), but would need to be adapted to include `test/Makefile`.

purcell:
I think you'd just add those flags [here](https://github.com/scfc/ledger-mode/blob/ignore-init-files-when-byte-compiling/test/Makefile#L2), right?

scfc:
@purcell Yes; I was hesitant because I did not understand `test-interactive`'s purpose, but (AFAICT) it does not interfere with that (I assume the warning about modified buffer `ledger-tests-26199BbX` is as designed?).  I'll post a pull request to see if Travis CI likes it.

jabranham:
The Makefile includes `--quick` which should take care of this issue now.

PR #97, opened 2018-02-20
Ignore init files when calling Emacs

scfc:
This change fixes issue #94.


scfc:
This does not load the autoloads for `flymake`, therefore failing that test.  I'll think about how to fix that while retaining the spirit of reproducible tests.

scfc:
Memo to self: With "does not load the autoloads for flymake, therefore failing that test", I apparently meant the second Travis CI job, for example https://travis-ci.org/ledger/ledger-mode/jobs/516855146, currently marked as "allowed failure".

jabranham:
The makefile seems to have --quick in it now, which should take care of this so I'm closing this PR.

PR #98, opened 2018-02-20
Setup bug-reference variables in .dir-locals

jabranham:


purcell:
Thanks. Why override the regexp here? Are there some formats used in the repo which aren't handled by the default? By adding this regexp, we're really saying that these are the linking styles we actively encourage, so I'd personally be very conservative here and stick to `#[0-9]+` if specifying explicit patterns.

jabranham:
I was trying to match what I saw in the repo's history already. So for
example when merging PRs, sometimes they're merged with a commit message
like:

Merge pull request #12 from joe

And othertimes, issues are referenced like so: Solves GH-12

purcell:
Cool, thanks.

PR #99, opened 2018-02-20
ledger-report: New option, ledger-report-resize-window

jabranham:
This controls whether ledger-report will call
shrink-window-if-larger-than-buffer or not.


purcell:
Thanks!

Issue #101, opened 2018-02-25
Issues detecting correct accounts


jabranham:
I'm having two different probably related issues:

1. ledger-magic-tab offers "assert commodity =='$'" when you define accounts like so at the top:

```
account assets:checking:
    assert commodity == "$"
```

So if you have (with | representing point):

```
2018-02-24
    ass|
```

and press TAB, you wind up with

```
2018-02-24
    assert commodity == "$"
```

2. When I do `M-x ledger-report`, and select a report that has `%(account)`, the resulting completions list to select an account includes non-accounts (for me, `; CSV: "Cleared", "01/06/2018", "restaurant-name"` is offered, for example).

Perhaps there should be a way to tell `ledger-mode` that the file in question uses e.g. `--pedantic` so that all the accounts are pre-declared at the top?

DamienCassou:
Please check if https://github.com/ledger/ledger-mode/pull/124 fixes the issue.

DamienCassou:
Is this issue fixed now?

jabranham:
Yup! Thanks.

PR #102, opened 2018-02-26
Allow pedantic account completion

jabranham:
Fixes bug #101

### DIFF
--- a/ledger-complete.el
+++ b/ledger-complete.el
@@ -44,6 +44,14 @@ This file will then be used as a source for account name completions."
   :type 'file
   :group 'ledger)
 
+(defcustom ledger-accounts-be-pedantic nil
+  "If non-nil, be pedantic in account completion.
+
+In other words, only offer accounts for completion that have been
+specified using an \"account\" directive."
+  :type 'boolean
+  :group 'ledger)
+
 (defun ledger-parse-arguments ()
   "Parse whitespace separated arguments in the current region."
   ;; this is more complex than it appears to need, so that it can work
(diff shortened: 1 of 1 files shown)


purcell:
I feel like this is only a workaround for #101, and not yet an actual fix, which would be preferable. I think the regexes used to identify accounts probably need to be tightened up, or scoped more carefully.

jabranham:
You're right, fix was a bad word choice on my part.

Is there a reason we don't just ask ledger for the accounts? Seems easier
than trying to figure out regexs....

jabranham:
This is along the lines of what I was thinking. (ignore the `(if nil ... ` part, I just kept it while testing it out:

```elisp
(defun ledger-accounts-list-in-buffer ()
  "Return a list of all known account names in the current buffer as strings.
Considers both accounts listed in postings and those declared with \"account\" directives."
  (save-excursion
    (goto-char (point-min))
    (let (results)
      (if nil (progn
                (while (re-search-forward (if ledger-accounts-be-pedantic
                                              ledger-account-directive-regex
                                            ledger-account-name-or-directive-regex) nil t)
                  (setq results (cons (match-string-no-properties 2) results))))
        (setq results (split-string
                       (shell-command-to-string
                        (concat "ledger -f " (buffer-file-name) " accounts" ))
                       "\n")))
      (sort (cl-delete-duplicates results :test 'string=)
            #'ledger-string-greaterp))))
```

purcell:
> Is there a reason we don't just ask ledger for the accounts? Seems easier than trying to figure out regexs....

Yes. `ledger accounts` doesn't list accounts declared with `account` directives, or in included files IIRC.

Re. the proposed change, I don't think the logic for finding accounts for completion purposes is modal in any way, so I feel like switching between two regexes based on a flag isn't the right approach.

jabranham:
OK, I'll close this then while we think of another way to solve the issue. 

People who want this fix now before a proper fix is out can use `(setq ledger-account-name-or-directive-regex ledger-account-directive-regex)`.

DamienCassou:
Please check if https://github.com/ledger/ledger-mode/pull/124 fixes the issue.

PR #104, opened 2018-03-17
Remove ledger-report-kill keybindings

jabranham:
The function isn't defined anywhere so these all produce errors. I
re-define the function as an obsolete alias for ledger-report-quit and
replace the keybindings.


purcell:
Great, thanks.

PR #105, opened 2018-03-17
Use sharp quotes in define-key calls

jabranham:
This lets the byte compiler catch issues like that in #104. It actually did another in this case - we were binding a key to `ledger-report-edit`, even though that's no longer a function. I updated it to point to `ledger-report-edit-report`.


purcell:
Ah, sorry, I merged #104 and now this conflicts. Would you mind rebasing/resolving this?

jabranham:
Done, thanks

purcell:
Fast, thanks! Merged.

PR #107, opened 2018-04-04
Make all faces inherited

jabranham:
This PR makes all faces inherited rather than use specific foregrounds. This should make ledger play nicely with most/all themes. I tested it out with some of the themes that ship with Emacs and it looks decent to me.


jabranham:
Just rebased off master. Any reason not to merge this?

jabranham:
No problem at all! Thanks!

PR #111, opened 2018-04-15
Fix a regexp in ledger-report--cmd-needs-links-p

cpitclaudel:
A command like 'ledger reg' does not include a final space, and thus wouldn't
get links.  This bug was surfaced by e098be20cf603f48dfe18c6237546a8c49e1f97c.


purcell:
Great, thanks.

PR #113, opened 2018-04-18
Add option that allows --strict to be used for reports

hexagonal-sun:
When running reports, add an option that allows a user to specify whether
--strict is passed to ledger when running reports.

### DIFF
--- a/ledger-report.el
+++ b/ledger-report.el
@@ -121,6 +121,12 @@ Calls `shrink-window-if-larger-than-buffer'."
   :type 'boolean
   :group 'ledger-report)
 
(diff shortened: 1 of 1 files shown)


purcell:
Thanks, this looks good, but could you rename `ledger-report-should-run-strict` to `ledger-report-use-strict` please? It's more consistent, and "should" is normally redundant.

purcell:
Thanks - the code still refers to `ledger-report-should-run-strict` though.

hexagonal-sun:
Oops, missed that.  Fixed now, apologies!

PR #114, opened 2018-05-07
schedule: maintain order of same date transactions in output

delexi:
Transaction that were scheduled on the same date had their order reversed in the
schedule buffer compared to their order in the schedule file.

This commit maintains the order defined in the schedule file.


delexi:
I can't really figure out what exactly breaks the travis build. It seems like the last step running `make checkdoc` fails for some reason. If I run that locally on my branch, the command's exit code is 0.

PR #117, opened 2018-06-04
Reconcile: Be aware of ellided default account postings

### LINKED ISSUE #58: Reconciliation mode doesn't work well with default accounts
`ledger-reconcile-visit` goes to the top of the ledger file and `ledger-reconcile-toggle` is a noop in a transaction where the account being reconciled is omitted as the default account.

delexi:
Fixes #58. See that issue for further detail.


purcell:
Great, thanks.

PR #119, opened 2018-06-11
Pass buffer instead of filename as BUFFER parameter

DamienCassou:
Current code assumes that the ledger configuration file will be opened
in a similarly named buffer. This can break. New code removes this
assumption and passes the correct buffer as parameter.


purcell:
Thanks!

PR #121, opened 2018-06-20
regex for effective date added

cacology:
A slight addition to the date regex so that date strings that include an effective date match.

This came up because I wanted to be able to run the "payee" report on transactions that have an effective date in addition to the transaction date.  I think it works, but it's worth looking over by someone with more experience with this code.

I'm happy to try again if it doesn't work.

Respectfully submitted,
-James

### DIFF
--- a/ledger-context.el
+++ b/ledger-context.el
@@ -39,7 +39,7 @@
 (defconst ledger-balance-assertion-string ledger-balance-assertion-regexp)
 (defconst ledger-comment-string "[ \t]*;[ \t]*\\(.*?\\)")
 (defconst ledger-nil-string "\\([ \t]+\\)")
-(defconst ledger-date-string "^\\([0-9]\\{4\\}[/-][01]?[0-9][/-][0123]?[0-9]\\)")
+(defconst ledger-date-string "^\\([0-9]\\{4\\}[/-][01]?[0-9][/-][0123]?[0-9]\\(?:=[0-9]\\{4\\}[/-][01]?[0-9][/-][0123]?[0-9]\\)?\\)")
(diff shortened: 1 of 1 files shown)


purcell:
Can you give a quick example to reproduce this please? There is `ledger-full-date-regex` which can already match optional effective dates, so there's some opportunity for consolidation in the code, but I'd need a bit more context.

cacology:
Thanks for your polite request!  I realize I'm not great at the *problem* and *solution* format.  Here's a minimal example of the problem:

# Problem

test.ledger
```
2019/01/02 New Dominion Bookstore  ; books
    Expenses:Reading:General  $10
    Liabilities:AMEX CC

2019/01/02=2019/01/15 New Dominion Bookstore  ; book deposit
    Expenses:Reading:General  $10
    Assets:Wells Fargo Bank:Checking
```

Placing the cursor on the first transaction, calling `ledger-report`, selecting `payee` works as expected.

Placing the cursor on the second transaction, after `...=2019/01/15 New Dominion`, calling `ledger-report`, selecting `payee`, throws an error:

```
Debugger entered--Lisp error: (wrong-type-argument stringp nil)
  regexp-quote(nil)
  ledger-report-payee-format-specifier()
  ledger-report-expand-format-specifiers("%(binary) -f %(ledger-file) reg @%(payee)")
  ledger-report-cmd("payee" nil)
  ledger-report("payee" nil)
  funcall-interactively(ledger-report "payee" nil)
  call-interactively(ledger-report nil nil)
  command-execute(ledger-report)
```

purcell:
Thanks! - I've committed a similar fix to this, re-using a regexp from `ledger-regex.el` to reduce duplication.

PR #124, opened 2018-07-20
Fix account-recognizing regular expressions

DamienCassou:
- exclude comments from being recognized as accounts
- remove [] and () from account names in virtual postings

Add a unit test to check that all account references and only those are detected.

To do the above, I had to fix ledger-post-align-postings which previously only working because comment lines were considered accounts.

### DIFF
--- a/ledger-post.el
+++ b/ledger-post.el
@@ -90,48 +90,35 @@ Looks only as far as END, if supplied, otherwise `point-max'."
 (defun ledger-post-align-postings (beg end)
   "Align all accounts and amounts between BEG and END, or the current region, or, if no region, the current line."
   (interactive "r")
-
-  (save-excursion
-    (let ((inhibit-modification-hooks t)
-          acct-start-column acct-end-column acct-adjust amt-width amt-adjust
-          (lines-left 1))
-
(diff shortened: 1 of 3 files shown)


jabranham:
Thanks, this seems to fix my issue. 

As a side note, it would be nice if ledger didn't suggest "closed" accounts. You can "close" an account in ledger by doing something like:

```
account OldChecking
    assert date <= [2018-01-01]
```

but that's outside the scope of this PR I guess :-)

Fuco1:
@DamienCassou  @jabranham sorry to hijack this thread but that "closing account" thing is very interesting.  I've asked about closing accounts on the IRC channel and got a bunch of responses I couldn't really evaluate :D Is there any place I can read more about the assertion here?

jabranham:
@Fuco1 they're underdocumented (there's an open issue on ledger's github to improve that), all the docs say is:

```
The check and assert directives warn or raise an error (respectively) if the given value expression evaluates to false within the context of any posting.
```

which you can find here: https://www.ledger-cli.org/3.0/doc/ledger3.html#Command-Directives

DamienCassou:
ping

PR #130, opened 2018-08-22
Silence the byte compiler

jabranham:
- Declare two variables used by multiline font-lock
- Require ledger-fontify in ledger-fonts for ledger-fontify-xact-state-overrides

### DIFF
--- a/ledger-fontify.el
+++ b/ledger-fontify.el
@@ -32,6 +32,10 @@
 (require 'ledger-regex)
 (require 'ledger-state)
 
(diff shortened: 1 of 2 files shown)


purcell:
Thanks. I'm going to invite you to be a committer here, since we have quite a liberal policy around this - please feel free to commit little fixes like this directly, or continue to file pull requests if you'd like some feedback before landing larger changes.

purcell:
Ugh, I can't figure out how to do this right now, sorry.

PR #132, opened 2018-08-28
 Add ledger-copy-transaction-insert-blank-line-after

Zaeph:
Hello,

I’ve written a small patch to add the possibility for user to insert a blank line after a transaction inserted with `ledger-copy-transaction-at-point`. The default behaviour only inserted a single `\n` after the transaction, which required manual fixing every time a transaction was to be copied.
I’ve also updated the manual to mention the variable in the relevant node (*Copying Transactions, Editing Amounts, Adding Transactions, The Ledger Buffer*).

I’m not sure if the Elisp is as good as it could be, nor if it flouts any convention I may not know, but I’m not very advanced with the language or with coding in general (although I do hope to improve).

(I’m still very new to the concept of contributing to projects and to opening pull requests, so please excuse me if I’ve done anything wrong, and thank you for bearing with me whilst I figure the ins and outs.)

### DIFF
--- a/ledger-xact.el
+++ b/ledger-xact.el
@@ -132,7 +132,7 @@ MOMENT is an encoded date"
          (encoded-date (ledger-parse-iso-date date)))
     (ledger-xact-find-slot encoded-date)
     (insert transaction
-            (if (bound-and-true-p ledger-copy-transaction-insert-blank-line-after)
+            (if ledger-copy-transaction-insert-blank-line-after
                 "\n\n"
               "\n"))
     (beginning-of-line -1)
--- a/ledger-xact.el
+++ b/ledger-xact.el
@@ -120,6 +120,9 @@ MOMENT is an encoded date"
                        mark desc)))))
       (forward-line))))
 
+(defvar ledger-copy-transaction-insert-blank-line-after nil
+  "Non-nil means insert blank line after a transaction inserted with ‘ledger-copy-transaction-at-point’.")
+
 (defun ledger-copy-transaction-at-point (date)
   "Ask for a new DATE and copy the transaction under point to that date.  Leave point on the first amount."
   (interactive  (list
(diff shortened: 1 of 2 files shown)


ledger-xact.el
```diff
@@ -128,7 +131,10 @@ MOMENT is an encoded date"
          (transaction (buffer-substring-no-properties (car extents) (cadr extents)))
          (encoded-date (ledger-parse-iso-date date)))
     (ledger-xact-find-slot encoded-date)
-    (insert transaction "\n")
+    (insert transaction
+            (if (bound-and-true-p ledger-copy-transaction-insert-blank-line-after)
```
purcell: Don't use `bound-and-true-p` here: if the var you've defined a few lines earlier isn't bound, there's a big problem! :-)

purcell:
I left a brief comment about a small fix that's needed. I do wonder if we should simply change the existing code to simply insert two newlines instead of one, though: everyone uses blank lines between transactions, don't they? Alternatively, we can make the `\n\n` or `\n` string the custom variable, rather than making it a boolean flag. That might be more straightforward and flexible, in some ways.

ledger-xact.el
Zaeph: Thank you for the feedback.

Maybe I have the wrong understanding of what a `bound` variable is, but wouldn’t `bound-and-true-p` cover the use we want for the variable, namely:
- Variable is initialised in `ledger-xact.el` with the value `nil`,
- Variable can be set to a `non-nil` value to enable the `\n\n` insertion.

I guess we don’t need to check whether the variable is bound, but I must have picked this up as a habit from previous forrays in Elisp.

Should I use this instead:
```diff
-(if (bound-and-true-p ledger-copy-transaction-insert-blank-line-after)
+(if ledger-copy-transaction-insert-blank-line-after
    "\n\n"
  "\n"))
```

Zaeph:
I would assume that everyone uses blank lines between transactions. I don’t remember seeing a single occurrence of the opposite, either in the manual or elsewhere. Regardless, I didn’t want to remove the original behaviour without allowing for a way to restore the original one with an option.

If you think that’s not warranted, I can indeed change `\n` to `\n\n`.

PR #135, opened 2018-09-28
Define pcomplete-uniquify-list if not defined

jabranham:
Emacs 27 fixes the typo and obsoletes the old function, so we'll use
defalias to define the typo-free function unless it's already bound.


purcell:
Thanks.

PR #136, opened 2018-09-30
ledger-report: New hook variable

jabranham:
`ledger-report-after-report-hook' is run after creating the
ledger-report buffer, inserting contents, and setting local variables
like `ledger-report-name'.

I find it useful to put a function in there that moves point to the
bottom of the report buffer but make it conditional on
`ledger-report-name'.

### DIFF
--- a/ledger-report.el
+++ b/ledger-report.el
@@ -128,6 +128,11 @@ when running reports?"
   :type 'boolean
   :group 'ledger-report)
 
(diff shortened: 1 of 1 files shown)


jabranham:
@purcell any chance of getting this and #138 reviewed or merged soonish?

purcell:
This seems uncontroversial, thanks. :-)

PR #137, opened 2018-10-02
Fix flymake version check

jabranham:


jabranham:
Otherwise it doesn't work in e.g. Emacs 27

purcell:
Thanks!

PR #138, opened 2018-10-04
ledger-report-toggle-default-commodity: New command

jabranham:
Bound to "$" in ledger-report-mode-map

### DIFF
--- a/ledger-report.el
+++ b/ledger-report.el
@@ -27,12 +27,15 @@
 
 (require 'ledger-xact)
 (require 'ledger-navigate)
+(require 'ledger-commodities)
 (declare-function ledger-read-string-with-default "ledger-mode" (prompt default))
 (declare-function ledger-read-account-with-prompt "ledger-mode" (prompt))
 
 (require 'easymenu)
 (require 'ansi-color)
 (require 'font-lock)
+(eval-when-compile
+  (require 'rx))
 
 (defvar ledger-buf)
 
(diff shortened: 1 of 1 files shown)


ledger-report.el
```diff
@@ -184,6 +187,7 @@ when running reports?"
       #'ledger-report-edit-report)
     (define-key map (kbd "M-p") #'ledger-report-previous-month)
     (define-key map (kbd "M-n") #'ledger-report-next-month)
+    (define-key map [?$] #'ledger-report-toggle-default-commodity)
```
purcell: For consistency, can I suggest:

```suggestion
    (define-key map (kbd "$") #'ledger-report-toggle-default-commodity)
```

ledger-report.el
```diff
@@ -643,6 +647,20 @@ arguments returned by `ledger-report--compute-extra-args'."
   (interactive)
   (ledger-report--change-month 1))
 
+(defun ledger-report-toggle-default-commodity ()
+  "Add or remove \"--exchange `ledger-reconcile-default-commodity' to the current report."
+  (interactive)
+  (unless (derived-mode-p 'ledger-report-mode)
+    (user-error "Not a ledger report buffer"))
+  (if (string-match
```
purcell: Here you probably want `string-match-p`, which avoids setting `match-string` etc. as a side-effect.

ledger-report.el
jabranham: Ok, changed.

ledger-report.el
jabranham: Done, thanks

jabranham:
Thanks @purcell I made those changes you recommended, rebased off master, and pushed here.

purcell:
Awesome, thanks!

PR #139, opened 2018-10-22
Fix ledger-report-toggle-default-commodity

jabranham:
@purcell turns out we need `string-match` because of the call to `replace-match`. So I added it back and wrapped it in `save-match-data`.


purcell:
Oh cool, thanks - sorry for the bad pointer. :-)

jabranham:
No problem - I really should've caught that since I wrote it!

PR #145, opened 2018-11-07
Make it easy to fetch ofx files and run ledger-autosync through them

DamienCassou:


DamienCassou:
Is this code of interest to anyone else than me? Is it something you would like to be part of ledger-mode or should I create a separate package?

/cc @jwiegley @yurikhan @jabranham @enderw88 @hexagonal-sun @delexi @cpitclaudel

cpitclaudel:
:clap: :clap: 
It's very interesting to me. Thanks a lot for posting this.
I think we should have it here, if you're happy to own the files for maintainer-ship purposes :)

yurikhan:
Omg. I visited the site mentioned in the commentary.

I choose to not be offended by that name and their application icons, but somebody most certainly will.

DamienCassou:
> I think we should have it here, if you're happy to own the files for maintainer-ship purposes :)

I will ask for commit right if we go that way because I want to be able to react fast if there is a problem.

> I choose to not be offended by that name and their application icons, but somebody most certainly will.

that's possible indeed and already happened IIRC. But I don't think that should be a problem for Ledger-mode.

ledger-import.el
```diff
@@ -0,0 +1,326 @@
...
+  :group 'ledger-import
+  :type '(repeat string))
+
+(defcustom ledger-import-boobank-command '("boobank")
+  "List of strings with boobank command name and arguments."
+  :group 'ledger-import
+  :type '(repeat string))
+
+(defcustom ledger-import-boobank-import-from-date "2018-10-01"
+  "String representing a date from which to import OFX data with boobank."
```
jabranham: Should probably specify what format(s) are acceptable for the date.

ledger-import.el
```diff
@@ -0,0 +1,326 @@
...
+  (interactive)
+  (pop-to-buffer-same-window (or buffer (ledger-import-buffer))))
+
+(defun ledger-import--non-empty-string-widget-matcher (_widget value)
+  "Return non-nil if VALUE is a non-empty string."
+  (and (stringp value)
+       (> (length value) 0)))
+
+(defun ledger-import-account-ledger-name (account)
+  "Return Ledger account name for ACCOUNT as known by your Ledger file."
```
jabranham: I don't understand this docstring. I guess ACCOUNT is assumed to have some structure? If so, the docstring should probably mention that.

ledger-import.el
```diff
@@ -0,0 +1,326 @@
...
+(defun ledger-import-account-fetcher-id (account)
+  "Return ACCOUNT identifier as known by the fetcher.
+For example, this is the account ID that boobank uses."
+  (nth 2 account))
+
+(defun ledger-import-account-fid (account)
+  "Return FID for ACCOUNT, or nil if none is necessary.
+This can be useful for ledger-autosync if the OFX data does not provide any."
+  (let ((fid (nth 3 account)))
+    (if (or (null fid) (string= fid ""))
```
jabranham: if (<cond>) nil can just be `unless`.

ledger-import.el
```diff
@@ -0,0 +1,326 @@
...
+      fid)))
+
+(defun ledger-import-choose-account ()
+  "Ask the user to choose an account among `ledger-import-accounts'."
+  (let* ((accounts ledger-import-accounts)
+         (account-name (completing-read "Ledger account: "
+                                        (mapcar #'ledger-import-account-ledger-name accounts)
+                                        nil
+                                        t)))
+    (seq-find (lambda (account) (string= (ledger-import-account-ledger-name account)
```
jabranham: Doesn't this need a `(require 'seq)`?

ledger-import.el
```diff
@@ -0,0 +1,326 @@
...
+(defun ledger-import-all-accounts (&optional ledger-file)
+  "Fetch transactions from all accounts and convert to Ledger format.
+Accounts are listed `ledger-import-accounts'.
+
+If LEDGER-FILE is non nil, use transactions from this file to
+guess related account names."
+  (interactive (list
+                (when (and (buffer-file-name) (derived-mode-p 'ledger-mode))
+                  (buffer-file-name))))
+  (require 'ledger-mode)
```
jabranham: This is unnecessary - it's already required at the top of this file.

ledger-import.el
```diff
@@ -0,0 +1,326 @@
...
+         (ledger-mode)
+         (ledger-mode-clean-buffer)
+         (run-hooks 'ledger-import-finished-hook)))
+     ledger-file)))
+
+(provide 'ledger-import)
+;;; ledger-import.el ends here
+
+;; Local Variables:
+;; eval: (flycheck-mode)
```
jabranham: Please no `eval` local variables, it makes opening the file annoying.

laurentb:
Hello,
The application is usually distributed through GPG signed packages (as are updates, which in addition go through https://updates.weboob.org/ by default). Passwords can be stored by tools like https://www.passwordstore.org/. No banking information goes through our websites, and the modules do verify https certificates.

We're migrating away from from Redmine (symlink.me) to Gitlab (git.weboob.org) so I suppose that was what you were referring to with using https correctly.

cpitclaudel:
> Omg. I visited the site mentioned in the commentary.

Ugh, gross.  Also, I think I misunderstood how this worked; maybe it should be an external package, after all? I thought the OFX parsing was done in Emacs, not by an external application.

ledger-import.el
DamienCassou: indeed. I can't wait for [Elsa](https://github.com/emacs-elsa/Elsa/) to be ready :-).

ledger-import.el
DamienCassou: I wondered the same. Apparently, it doesn't, the byte compiler is happy:

```session
$ emacs -L . -Q --batch --funcall batch-byte-compile ledger-import.el
$ echo $?
0
```

I added the `require` line anyway.

ledger-import.el
DamienCassou: indeed

ledger-import.el
DamienCassou: you are right

purcell:
I think I'd vote in favour of a separate package for this one.

PR #155, opened 2019-02-11
Fixed flymake backend error

plattfot:
The `make-process` commands only accepts a list of strings but when
`ledger-flymake-be-pedantic` and/or `ledger-flymake-be-explicit` is not set to
true it will fail as that list now contains nil. Changed it to just return an
empty string when a setting isn't enabled.

Edit: Fixed quotes in markdown

### DIFF
--- a/ledger-flymake.el
+++ b/ledger-flymake.el
@@ -72,8 +72,8 @@ Flymake calls this with REPORT-FN as needed."
         :name "ledger-flymake" :noquery t :connection-type 'pipe
         :buffer (generate-new-buffer " *ledger-flymake*")
         :command `(,ledger-binary-path "-f" ,file
-                                       ,(when ledger-flymake-be-pedantic "--pedantic")
-                                       ,(when ledger-flymake-be-explicit "--explicit")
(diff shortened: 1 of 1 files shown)


purcell:
I've just merged this, but does an empty string also cause issues? It'd mean extra (empty) args passed to the program, which might confuse it.

plattfot:
Hmm, good question. Tested:
```elisp
(start-process "ledger" "ledger" "ledger" "-f" "ledger.dat" "" "" "balance")
```
In the `ledger` buffer I see
```
Error: Unrecognized command ''

Process ledger<1> exited abnormally with code 1
```
It looks like it gets confused. So not sure why it works for the backend. Maybe safer to do something like this instead:

```elisp
(cl-remove nil `(,ledger-binary-path "-f" ,file
                                      ,(when ledger-flymake-be-pedantic "--pedantic")
                                      ,(when ledger-flymake-be-explicit "--explicit")
                                      "balance"))
```

purcell:
Ah! Would you mind filing another quick PR please?

PR #157, opened 2019-02-11
Cleaner fix for the flymake backend.

plattfot:
Removing the `nil` entries in the command list instead of setting them to empty
strings. As empty strings might mess with the command.

This cleaner and safer compared to the fix in #155.


purcell:
Thanks @plattfot!

PR #161, opened 2019-03-06
Added support to align amounts that are an expression.

plattfot:
Will pick the last value to align against.

See #158


enderw88:
This looks OK, but everytime someone touches these regexes, something else breaks.  I will try to test this shortly.

PR #177, opened 2019-06-14
Remove blank lines after ledger-delete-current-transaction

jabranham:
Otherwise when we delete a transaction between two transactions, we're
left with a few blank lines instead of just one.

### DIFF
--- a/ledger-xact.el
+++ b/ledger-xact.el
@@ -161,7 +161,8 @@ MOMENT is an encoded date"
   "Delete the transaction surrounging POS."
   (interactive "d")
   (let ((bounds (ledger-navigate-find-xact-extents pos)))
-    (delete-region (car bounds) (cadr bounds))))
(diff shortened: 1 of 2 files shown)


purcell:
I think it's legal to have no blank line between transactions: does this still work sanely in that case? The docstring for `delete-blank-lines` makes me wonder.

jabranham:
You mean if you have three transactions with no blank lines between them and you do `C-c C-d` on the middle one? The behavior there is unchanged.

purcell:
> You mean if you have three transactions with no blank lines between them and you do `C-c C-d` on the middle one?

Yep exactly.

> The behavior there is unchanged.

Sounds good, then! But the test `ledger-reconcile/test-027` is failing, and that's due to the change in `ledger-delete-current-transaction`, afaict.

jabranham:
Fixed, thanks

purcell:
Great, thank you!

PR #179, opened 2019-06-18
Make ledger-default-date-format a user option

### LINKED ISSUE #173: Should 'ledger-default-date-format' be a defcustom instead of a defvar?
I'm not hugely familiar with defcustom customs, but `ledger-default-date-format` seems like the kind of thing it was designed for. It's a user option, not merely an internal variable.

It looks like an easy change, just copying what `ledger-reconcile-default-date-format` does:

https://github.com/ledger/ledger-mode/blob/db42e8343166a38fb3b8c414009a444d42af0e58/ledger-reconcile.el#L78-L82

But maybe there's a reason defvar was chosen?

jabranham:
Closes #173


purcell:
Thanks. Worth noting in the docs?

jabranham:
Probably! I added a short sentence to the manual, let me know if you think something else is needed.

purcell:
Great, thank you.

PR #182, opened 2019-07-01
Don't kill the report buffer on every new report

jabranham:
Erase the report buffer rather than killing it and recreating it. This
allows users to use functions like `display-buffer-reuse-window' to
manager where the ledger-report buffer shows up.


purcell:
Makes sense, thanks.

PR #183, opened 2019-07-01
Match account names more liberally for the sake of completion

purcell:
See #141


purcell:
Can't reproduce this test failure locally, grrr.

PR #184, opened 2019-07-02
Don't include current line's payee in ledger-payees-in-buffer

### LINKED ISSUE #181: Tab completion broken for payee
Hi,

updated to ledger-mode to `20190619.314` today. I use company-mode for the completion. After the update, completion stopped working entirely. Reading around on the issues and the docs I discovered I needed to add this to my config
```
(add-hook 'ledger-mode-hook
                 (lambda ()
                   (setq-local tab-always-indent 'complete)
                   (setq-local completion-ignore-case t)
                   (setq-local ledger-complete-in-steps t)))
```
Also found out why some of my accounts stopped working with completion awhile back (#141).

The issue I have now is that it includes the payee I'm currently typing in the list of suggested completions. So I cannot complete the payee I actually want, without navigating down in the completion list company gives me.
 
For example
```
2019/06/28 Foobar
    Expenses:Baz                               11.99 CAD
    Assets:Cash

2019/06/20 Foo|
``` 
Where `|` is my cursor. If I hit tab I get Foo and Foobar, if I continue typing it will suggest Foob and Foobar and so on. Expected behavior is to just get Foobar.

I see the same behavior if I disable company. Only then I cannot complete the payee at all. As it always matches against what I'm typing.

jabranham:
Otherwise this makes payee completion difficult.

Closes #181

### DIFF
--- a/ledger-complete.el
+++ b/ledger-complete.el
@@ -82,15 +82,20 @@ If nil, full account names are offered for completion."
 
 
 (defun ledger-payees-in-buffer ()
-  "Scan buffer and return list of all payees."
+  "Scan buffer and return list of all payees.
+Omit the payee on the current line if it is the only instance of
+that payee in the buffer."
+  ;; We omit the payee on the current line because this function is part of
+  ;; `ledger-complete'. If we include the current line's payee, then anything
+  ;; matches, which makes it hard to do completion, see bug#181.
   (let ((origin (point))
         payees-list)
     (save-excursion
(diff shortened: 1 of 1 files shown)


purcell:
Makes sense, will merge, thanks. I wonder if these 'ledger-*-in-buffer` functions should be renamed to `ledger-complete-*-in-buffer`, since the naming implies they are general-purpose functions, which isn't really the case given tweaks like this.

PR #190, opened 2019-08-03
New commands ledger-navigate-{next/previous}-uncleared

jabranham:
These are two commands I use quite often to help me navigate between uncleared transactions. If people think they're useful enough to include in ledger-mode, I'm happy to write up documentation and some tests.


jabranham:
I added a test and some documentation, will merge this soonish unless someone objects.

PR #191, opened 2019-08-14
Add a way to filter account list for completion

jabranham:
New option ledger-accounts-exclude-function if non-nil is a function that can be used to filter accounts that ledger-mode offers for completion. It gets one argument which is a list containing the account name and any subdirectives associated with the account.

I made this because I have quite a few "closed" accounts that I don't want offered by completion:

```
account Assets:Old Checking
    assert date<=[2018-01-01]
```

There's also a new function ledger-accounts-in-buffer that returns a list of accounts along with any subdirectives.

### DIFF
--- a/NEWS.org
+++ b/NEWS.org
@@ -41,3 +41,7 @@ how to use this.
 New commands ledger-navigate-next-uncleared and
 ledger-navigate-previous-uncleared move to the next and previous uncleared
 transactions. These don't have default keybindings.
+** New option ledger-accounts-exclude-function
+This allows users to exclude some accounts from being offered during completion.
+This is useful if, for example, you have a lot of accounts but only use some of
+them rarely.
--- a/ledger-complete.el
+++ b/ledger-complete.el
@@ -90,15 +90,66 @@ If nil, full account names are offered for completion."
     ;; to the list
     (sort (delete-dups payees-list) #'string-lessp)))
 
+(defun ledger-accounts-in-buffer ()
+  "Return an alist of accounts in the current buffer.
+The `car' of each element is the account name and the `cdr' is an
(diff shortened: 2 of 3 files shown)


purcell:
Does this mean `ledger-accounts-list` is now unused. Bit confused now that there's a third function returning this sort of info.

jabranham:
No, it's still used. `ledger-accounts-list` checks the users `ledger-accounts-file` if set. `ledger-accounts-in-buffer`only considers the current buffer. `ledger-accounts-list-in-buffer` is similar to `ledger-accounts-in-buffer` but returns a simpler structure. We could remove that function entirely pretty easily but I didn't want to break external code.

kanreki:
Hi there,

I see that you added this feature in 90c7feb, a few years ago, and included a new test "ledger-complete/test-ledger-accounts-exclude-function".

As I was trying to run tests, I found that this test is failing. And it even is failing when I wind the Git repo back to your original commit. So, I'm wondering, did this test work for you? How did you get it to pass?

The new code in function `ledger-accounts-in-buffer` seems obviously troubling, since a directive can look like an account name. (We would need to be more selective in parsing past `account` directives that have "sub-directives" such as `assert`, it seems.)

```
F ledger-complete/test-ledger-accounts-exclude-function
    (ert-test-failed
     ((should
       (equal
	(ledger-accounts-list-in-buffer)
	(list "Assets:Checking:Bank B")))
      :form
      (equal
       ("Assets:Checking:Bank B" "assert date<=")
       ("Assets:Checking:Bank B"))
      :value nil :explanation
      (proper-lists-of-different-length 2 1
					("Assets:Checking:Bank B" "assert date<=")
					("Assets:Checking:Bank B")
					first-mismatch-at 1)))
```

PR #193, opened 2019-08-27
Ability to use single or multiple words for a ledger-exec-binary

evgeniysharapov:
With this  change in place one could use `wsl ledger` in place of a `ledger-binary-exec` and have it working on Windows given it installed in WSL.


evgeniysharapov:
Then you are probably better off by adding it to the PATH. In case of WSL,
you can 't just fiddle with environment variable, you have to ran it as a
`wsl ledger`.

PR #195, opened 2019-09-20
Read directory-local variables in reports

DamienCassou:
This is useful when a directory has a definition for the
variable`ledger-reports`. I guess it can be useful for other variables
too so I when for a simple and generic solution.

* ledger-report.el (ledger-report-mode): Call
`hack-dir-local-variables-non-file-buffer` so that directory-local
variables are taken into account.


jabranham:
Thanks, pushed as 1d229f88b57da8e3f9dda8f5ede2293126b29d55

PR #203, opened 2019-10-05
Remove ledger-report-select-report

jabranham:
This function doesn't seem to work (get a type error) and AFAICS its functionality is just `C-c C-o C-r` in ledger-report buffers. OK to remove?

### DIFF
--- a/ledger-report.el
+++ b/ledger-report.el
@@ -180,7 +180,6 @@ See documentation for the function `ledger-master-file'")
     (define-key map [?r] #'ledger-report-redo)
     (define-key map [(shift ?r)] #'ledger-report-reverse-report)
     (define-key map [?s] #'ledger-report-save)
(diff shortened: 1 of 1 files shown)


bcc32:
What's the type error you get?  It works for me, using the latest elpa version.

Also, this PR removes the menu item for `ledger-report-save`, which looks like a mistake.

jabranham:
> What's the type error you get? It works for me, using the latest elpa version.

Ah, it works if run from the report buffer, not from the actual ledger file.

> Also, this PR removes the menu item for ledger-report-save, which looks like a mistake.

Yes, that's a typo :-)

Looking at this more, it probably makes more sense to just mark `ledger-report-select` as an (obsolete) alias of `ledger-report` now that `ledger-report` works from report buffers.

Issue #194, opened 2019-09-04
please tag a release


bremner:
It looks like it's been 3 years and about 300 commits since the last release was tagged.

jabranham:
I think this is because ledger-mode itself has never had a release. Those tags are from when ledger and ledger-mode lived in the same repo. Seems reasonable to me to cut a new release though.

jabranham:
Who has the authority to make this happen? @purcell @enderw88 ?

purcell:
Any of us can do it. I'm happy to add a new tag, but that's hardly a release process. :-)

jabranham:
AFAIU ledger-mode hasn't ever had a release on its own so we probably don't have a release process. Seems like it's as good a time as any to come up with one. I've got no big ideas here, other than checking that the tests pass & there aren't compilation warnings. I suppose we could go through and appease checkdoc as well...

We also need to come up with a version number. `v3.1.1` is the last tag. I think probably the biggest change is the switch to how completion works, not sure if that merits `v4.0.0` or `v3.2.0`.

purcell:
Yeah, just wondering about a changelog, because there *will* be people using the super-old version in MELPA Stable.

jabranham:
Good idea. I can look through the log and put together some sort of a news
file.

jabranham:
@purcell @enderw88 I think we're OK to tag `v4.0.0`, right?

bitti:
Yeah, it's a shame that it's been so long. Is something else holding this off?

purcell:
Done now. MELPA Stable should build v4.0.0 in the next few hours.

PR #204, opened 2019-10-05
Add a news file

jabranham:
Cobbled together by glancing through the commits since the v3.1.1 tag. Very possible I missed something.

Hopefully we can tag a new release shortly after this gets merged (#194) 

@purcell


purcell:
Nice.

PR #205, opened 2019-10-06
Use nix emacs ci

purcell:
Here I've reworked the travis build to use emacs binaries from my `nix-emacs-ci` project, such that installation is more convenient and consistent. MacOS is still supported in addition to Linux, though - as before - MacOS builds are disabled.

Additionally, `ledger` is installed via Nix: stable versions come from nixpkgs unstable - these should be the latest releases afaik - and snapshot versions are built from source by nix.

I'll go ahead and merge this right away but wanted to leave a note for the other committers.

I may also investigate switching us to Github Actions, which is faster that Travis.


.travis.yml
```diff
@@ -1,26 +1,41 @@
-sudo: required
-language: generic
-dist: xenial
+language: nix
+
+cache:
+  directories:
+    - ledger-master
+    - ledger-next
```
jabranham: Are these lines necessary anymore @purcell? Unless I'm missing something they don't do anything?

purcell: No, they can be removed. I'll do that now.

PR #206, opened 2019-10-07
Try github actions

purcell:
Here's how a basic github actions setup might work. I don't think the feature is available yet for this organisation, though, so I'll file this as a PR for later reference.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,39 @@
+name: CI
+
+on:
+  push:
+    paths-ignore:
+    - '**/*.md'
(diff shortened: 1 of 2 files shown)


jabranham:
@purcell I think we have GitHub Actions now.

purcell:
Nix is a bit broken on MacOS right now, so holding off on this. Seems like the PR would need to be merged in order to activate Actions, since pushing commits to this branch doesn't do the trick.

purcell:
Now merged and made the default CI. Tests on both Linux and MacOS.

PR #212, opened 2019-11-09
Handle prefixed negation on commoditized amounts

### LINKED ISSUE #211: Context at point doesn't handle negative commoditized amounts properly
This seems to be because `ledger-commoditized-amount-regexp` doesn't properly handle negated amounts like `-$100`, so `ledger-extract-context-info` returns the account context with the separator and amount included in the account name.

For example calling `ledger-context-at-point` with point at `|` below:
```
2011/01/27 Book Store
    Expenses:Books                            $20.00
    Liabilities:MasterCard                   -$20.00|
```
returns:
`(acct-transaction account ((indent "   " 76) (status nil nil) (account "Liabilities:MasterCard                   -$20.00" 80)))`

I believe I have a fix and will open a PR shortly.

kevinjfoley:
This includes prefixed `-` characters when matching commoditized amounts.

Fixes #211

### DIFF
--- a/ledger-regex.el
+++ b/ledger-regex.el
@@ -299,13 +299,22 @@
 (ledger-define-regexp commoditized-amount
   (macroexpand
    `(rx (group
-         (or (and (regexp ,ledger-commodity-no-group-regexp)
-                  (*? blank)
-                  (regexp ,ledger-amount-no-group-regexp))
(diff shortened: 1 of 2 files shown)


bcc32:
Thanks for your contribution (from 2019!).  Rebased and merged with a small typo fix.

PR #221, opened 2020-02-11
Add the Yen symbol to ledger-amount-regex

samdoshi:
Add the Yen (¥) symbol to `ledger-amount-regex`.

Used for Japanese currency, might also be used for Chinese Yuan/Renminbi but I don't know enough about that...


purcell:
Thanks for this. I've committed a change in d619ba01 that supports all allowable commodity symbols.

PR #222, opened 2020-02-22
Use :extend in ledger-font-xact-highlight-face in Emacs>=27

jabranham:
This is a new face attribute in Emacs 27 and without this, the
highlighting stops at the end of the text instead of continuing to the
end of the window.


purcell:
Great, thanks.

PR #225, opened 2020-03-12
Don't require separator before comment

glguy:
ledger-mode is incorrectly requiring two spaces to precede comments, this PR resolves that. I'd like to note that while most of the examples and tests show a two-space separator before the comments that there exist ledger tests that only have one space before the comment and that ledger itself requires no spaces.


purcell:
Makes sense, thanks for this.

PR #227, opened 2020-03-27
Mark report buffer not modified in ledger-report-redo

### LINKED ISSUE #226: Report buffer considered modified after ledger-report-redo, prompts to save
After triggering `ledger-report-redo` by modifying the ledger file, the ledger-report buffer has its modified flag set.  When selecting another report with the `S` key binding, the user is prompted unnecessarily to save the buffer.

To reproduce:

1. `ledger-report` and select the default `bal` report.
2. Switch back to the buffer editing the ledger file and make some change and save the buffer, causing the report buffer to update.
3. Switch back to the report buffer and press `S` to select another report.
4. The user is prompted: "Buffer modified, save it? (y or n)".

I think the user should not be prompted, or otherwise the buffer should be marked as not modified by `ledger-report-redo` like `ledger-report` does.  I'm going to open a PR to do the latter, which seems somewhat better to me.

bcc32:
Fix #226.

An alternative would be to check at the beginning of `ledger-report` to see if the current buffer is a `ledger-report-mode` buffer and not prompt the user to save, but that seems less clean than this approach.

### DIFF
--- a/ledger-report.el
+++ b/ledger-report.el
@@ -576,6 +576,7 @@ arguments returned by `ledger-report--compute-extra-args'."
         (ledger-report-reverse-lines))
       (when ledger-report-auto-refresh-sticky-cursor
         (forward-line (- ledger-report-cursor-line-number 5)))
+      (set-buffer-modified-p nil)
       (run-hooks 'ledger-report-after-report-hook)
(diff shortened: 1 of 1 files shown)


jabranham:
Makes sense to me, thanks

PR #231, opened 2020-04-01
Mark report buffers not modified

bcc32:
I missed these in #227.  All locations where `inhibit-read-only` is bound are
now accounted for.

### DIFF
--- a/ledger-report.el
+++ b/ledger-report.el
@@ -168,14 +168,13 @@ See documentation for the function `ledger-master-file'")
 
 (defun ledger-report-reverse-lines ()
   "Reverse the lines in the ledger report buffer."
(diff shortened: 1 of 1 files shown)


bcc32:
Bump!  Any thoughts on this PR?  I think it's pretty straightforward.

purcell:
Thanks for the bump.

This looks fine, but I feel like all these instances should ideally just use the `with-silent-modifications` macro. Alternatively, or perhaps additionally, the report buffers should likely be made read-only.

bcc32:
Ah, I was not aware of that macro, thanks!  The report buffers are already marked read-only, AFAIK.

purcell:
Awesome, thank you!

PR #236, opened 2020-04-20
Avoid quadratic-time completion algorithm.

riastradh:
Use a hash table to avoid gathering duplicate account names, rather
than cl-pushnew which runs in O(n) time for each account.


purcell:
Thanks for this. I think #236 is a more direct equivalent, since it doesn't actually matter whether we keep entries with/without data, IIRC. Not even sure the data is used...

purcell:
Your fix is the right one - merged.

Issue #71, opened 2018-01-22
Highlighting issues on 'account' lines


cpitclaudel:
The ledger-font-xact-highlight-face overlay covers one character too much when positioned on account lines (maybe it should just be removed when the point isn't on a transaction?)

![ledger-mode-](https://user-images.githubusercontent.com/8181630/35228355-8440e0a4-ff5e-11e7-805f-4caf1f2d0a5f.png)

Additionally, comments on account lines are not highlighted as comments:

![ledger-mode](https://user-images.githubusercontent.com/8181630/35228244-3fc9a50a-ff5e-11e7-879e-e3bf30696f4e.png)

purcell:
Part 1 of this is fixed. The comment highlighting is pretty broken though. :-(

riastradh:
The first part of this still (again?) seems to be happening for me.

Versions:
- GNU Emacs 26.3
- ledger-mode 6a54eb2b762dd2a5a6ceff54019eeb430562948a

PR #238, opened 2020-04-20
Stop at end of last line, not at first character of next line.

riastradh:
Candidate fix for first part of <https://github.com/ledger/ledger-mode/issues/71>.


purcell:
Yes, looks like things changed after my previous fix. Thanks for this.

PR #239, opened 2020-04-20
Determine account completions faster in large files

purcell:
See #236, thanks @riastradh


purcell:
Ah no, will just keep the fix in #236...

PR #240, opened 2020-04-29
Add customization for adding `--explicit` flag to reports

yyadavalli:


yyadavalli:
No longer useful after ledger v3.2.0

purcell:
Cool, thanks for following up with your conclusion.

PR #241, opened 2020-05-03
update docs for the Ledger Post customization group

simonmichael:
Update to match the current four variables' names and help.

### DIFF
--- a/doc/ledger-mode.texi
+++ b/doc/ledger-mode.texi
@@ -1146,11 +1146,20 @@ Ledger Post:
 
 @ftable @option
 
-@item ledger-post-amount-alignment-column
-The column Ledger-mode uses to align amounts.
+@item ledger-post-account-alignment-column
+The column Ledger-mode attempts to align accounts to.
 
-@item ledger-default-acct-transaction-indent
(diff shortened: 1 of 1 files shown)


doc/ledger-mode.texi
```diff
@@ -1146,11 +1146,21 @@ Ledger Post:
...
 
+@item ledger-post-account-alignment-column
+The column Ledger-mode attempts to align accounts to.
+
+@item ledger-post-amount-alignment-at
+Position at which the amount is aligned.
+
+Can be :end to align on the last number of the amount (can be
+followed by unaligned commodity) or :decimal to align at the
+decimal separator.
```
basil-conto, line 1157: Should `:end` and `:decimal` be marked as `@code{}`?

simonmichael: Thanks, I've re-pushed with that change.

purcell:
Thanks!

simonmichael:
@purcell: just a reminder that https://www.ledger-cli.org/3.0/doc/ledger-mode.html#Ledger-Post-Customization-Group needs to be re-rendered.

Issue #237, opened 2020-04-20
Completion no longer goes component by component


riastradh:
Suppose I have the following accounts ([example ledger](https://github.com/ledger/ledger-mode/files/4500610/test.txt)):

- Assets:Bank:Deposit:20200419
- Assets:Bank:Balance
- Assets:Reimbursements

It used to be that when I typed `As<TAB>`, it would expand to `Assets:`, and then if I typed `B<TAB>` it would expand to `Assets:Bank:`, and then if I typed `D<TAB>` it would expand to `Assets:Bank:Deposit:`.  This has been trained in my muscle memory for years.

Now when I type `As<TAB>`, nothing happens.  The [documentation](https://github.com/ledger/ledger-mode/blob/4dd9b65184fbdac2499368160e9eee5732788399/doc/ledger-mode.texi#L301-L305) suggests that I do

```
(setq-local tab-always-indent 'complete)
(setq-local completion-cycle-threshold t)
(setq-local ledger-complete-in-steps t)
```

But if I do that, then when I type `As<TAB>`, it immediately expands to `Assets:Bank:` even though `Assets:Reimbursements` is another possible completion.

I don't want to disable cycling -- I do want it to cycle through each expansion of each component in the account hierarchy, so that, e.g., after `Assets:Bank:` it will cycle between `Assets:Bank:Balance` and `Assets:Bank:Deposit:`.  But it shouldn't skip ahead and ignore possible completions like `Assets:Reimbursements` -- and I would prefer that it restore the old behaviour of never completing more than one component at a time even if unambiguous, so that I can reliably hit `<TAB>` to complete at most one component as I did before.

Versions:
- GNU Emacs 26.3
- ledger-mode 6a54eb2b762dd2a5a6ceff54019eeb430562948a

purcell:
I just tried this out in Emacs 26.3 and it worked as you wanted it to. Perhaps there are other completion settings affecting the behaviour. What completion system are you using?

riastradh:
```
% git status
On branch master
Your branch is up to date with 'origin/master'.

nothing to commit, working tree clean
% cd ..
% mkdir build-master
% cd build-master
% time nice -n +10 env -i PATH=/bin:/usr/bin:$HOME/usr/bin cmake ../master
[...]
% time nice -n +10 env -i PATH=/bin:/usr/bin:$HOME/usr/bin make -j12
[...]
% emacs -nw -q

; *scratch* buffer
[...]

PR #242, opened 2020-05-09
Use a cache for account completion

jabranham:
Fix for #237. 

Requires `completion-table-with-cache` which requires Emacs 24.4 (we currently support back to 24.3). I took the opportunity to bump our minimum version up to 25. 25 was released ~4 years ago and AFAIK all the major distros have updated to at least 25.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -14,7 +14,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest]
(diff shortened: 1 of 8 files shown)


jabranham:
Anyone object to dropping support for Emacs 24 in order to merge this fix? 

@purcell @jwiegley @enderw88 (sorry if I forgot anyone else...)

purcell:
No objection from me.

PR #248, opened 2020-05-27
Switch to newer, faster install-nix-action

purcell:


purcell:
Pending https://github.com/ledger/ledger/pull/1919

purcell:
> ok to merge?

Yes, done now.

PR #255, opened 2020-06-22
Transaction folding

drdv:
Even though `ledger-occur` can be used to narrow down a buffer based on a regex, I find it useful to have, in addiction, folding of transaction records. This PR implements the latter. For example, folding a transaction
```
2011/01/02 Grocery Store
    Expenses:Food:Groceries                  $ 65.00
    * Assets:Checking
```
results in
```
2011/01/02 Grocery Store...
```
Two `interactive` functions are exposed:
- `ledger-mode-transaction-toggle-folding`: hide/show transaction under point 
- `ledger-mode-folding-toggle-transactions`: hide/show all transactions in the buffer

`ledger-occur` and transaction folding can be used to complement each other.

The buffer-local variable `ledger-mode-toggle-invisible-transactions` can be used to enable/disable folding of invisible transactions (i.e., transactions whose overlay has been set using `ledger-occur`).

Transaction blocks are detected using the same approach as in `ledger-highlight-xact-under-point` (probably some code refactoring should be done).

The transaction folding functionality has been added to the `ledger-occur.el` file even though probably a new file should be used (my aim was to do the least amount modifications). The same thing applies to the unit tests. 

The function `ledger-occur-make-invisible-overlay` has been modified by adding `(overlay-put ovl 'display "")` in order to avoid undesired trailing ellipsis when using `ledger-occur` together with transaction folding. 

Overlay of type `'code` has been used in order for incremental search to temporarily open folded blocks.

### DIFF
--- a/ledger-occur.el
+++ b/ledger-occur.el
@@ -202,13 +202,11 @@ temporarily opened when doing incremental search."
         (when (and (> (- e b) 1)            ; not an empty line
                    (<= p e) (>= p b)        ; point is within the boundaries
                    (not (region-active-p))) ; no active region
-          (goto-char b)
           (save-excursion
+            (goto-char b)
             (goto-char (line-end-position))
             (if (hs-overlay-at (point))  ;; if transaction is hidden show it
-                (progn
-                  (save-excursion (hs-show-block))
-                  (goto-char b))
+                (hs-show-block)
               (goto-char b)
               (hs-discard-overlays (line-end-position) e)
               (hs-make-overlay (line-end-position) e 'code)
--- a/ledger-occur.el
+++ b/ledger-occur.el
@@ -193,7 +193,7 @@ A transaction block is identified as in ledger-highlight-xact-under-point.
 Overlay of type `code' is used so that hidden blocks are
 temporarily opened when doing incremental search."
   (interactive)
-  (if (not (and (boundp 'hs-minor-mode) hs-minor-mode))
+  (if (not (bound-and-true-p hs-minor-mode))
       (message "Enable hs-minor-mode to use this functionality.")
     (let ((exts (ledger-navigate-find-element-extents (point))))
       (let ((b (car exts))
--- a/ledger-occur.el
+++ b/ledger-occur.el
@@ -31,6 +31,7 @@
 
 (require 'cl-lib)
 (require 'ledger-navigate)
+(require 'hideshow)
 
 (defconst ledger-occur-overlay-property-name 'ledger-occur-custom-buffer-grep)
 
@@ -119,7 +120,9 @@ long, otherwise it is the word at point."
   "Make an overlay for an invisible portion of the buffer, from BEG to END."
   (let ((ovl (make-overlay beg end)))
     (overlay-put ovl ledger-occur-overlay-property-name t)
-    (overlay-put ovl 'invisible t)))
+    (overlay-put ovl 'invisible t)
+    ;; required to not display ... when using together with transaction folding
+    (overlay-put ovl 'display "")))
 
 (defun ledger-occur-create-overlays (ovl-bounds)
   "Create the overlays for the visible transactions.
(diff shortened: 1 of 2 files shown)


drdv:
I am not sure what is the best way to handle the CI build error. Currently, the user might choose not to activate `hs-minor-mode` if he/she doesn't want to use the folding functionality. On the other hand, a warning is displayed when attempting to use the folding functionality without enabling `hs-minor-mode`. So far I haven't experienced problems with this setup. 

Any suggestions are welcome.

josephmturner:
I'm just stumbling across this now, and I like this idea. Is there a reason why this PR was not merged? Currently, I'm using the simpler folding technique described [here](https://github.com/simonmichael/hledger/issues/367#issuecomment-232492323):
```
; M-1 to collapse all transactions to one line, M-0 to reset. Useful for quickly scanning
(global-set-key "\M-0" (lambda () (interactive) (set-selective-display (* tab-width 0))))
(global-set-key "\M-1" (lambda () (interactive) (set-selective-display (* tab-width 1))))
(add-hook 'ledger-mode-hook (lambda () (setq tab-width 4)))
```

drdv:
I don't know why it is not merged.

drdv:
I just noticed that this PR is still pending. Is there anything I can do to help? I haven't followed the development for a while ... maybe this feature is no longer interesting?

bcc32:
I started playing around with this.  Seems useful, but I have a couple questions about the implementation:

Why define separate commands for toggling hide/show specifically for ledger-mode, when hideshow already defines such commands?  Couldn't we just configure `hs-find-block-beginning-func` and the other corresponding functions for ledger-mode and then users could invoke the usual `hs-*` commands?

Just to confirm: if I'm understanding the code correctly, it is using hideshow's functionality to provide some overlay-related helpers and also things like the `'code` overlay type (to support isearch temporarily displaying the contents).  But it doesn't really use `hs-minor-mode` except for the invisibility spec, right?

ledger-occur.el
```diff
@@ -173,6 +176,76 @@ BUFFER-MATCHES should be a list of (BEG END) lists."
...
 
+;; -----------------------------------------------------------------------------
+;; transactions folding
+;; FIXME: probably this is not the best file to include this functionality.
+;; -----------------------------------------------------------------------------
+(defvar-local ledger-mode-folding-transactions-hidden nil
+  "Whether transactions are globally hidden or not.")
+
+(defvar-local ledger-mode-toggle-invisible-transactions t
+  "Toggle invisible transactions (see ledger-occur-mode).")
```
bcc32, line 187: This is a normal variable, not a user customization, but nothing in the code sets its value.  Is that intentional?

ledger-occur.el
```diff
@@ -173,6 +176,76 @@ BUFFER-MATCHES should be a list of (BEG END) lists."
...
+
+(defun ledger-mode-transaction-toggle-folding ()
+  "Toggle hiding of transaction block under point.
+A transaction block is identified as in ledger-highlight-xact-under-point.
+
+Overlay of type `code' is used so that hidden blocks are
+temporarily opened when doing incremental search."
+  (interactive)
+  (if (not (and (boundp 'hs-minor-mode) hs-minor-mode))
```
bcc32, line 196: BTW this can be simplified by using the `bound-and-true-p` macro.

ledger-occur.el
```diff
@@ -173,6 +176,76 @@ BUFFER-MATCHES should be a list of (BEG END) lists."
...
+
+(defun ledger-mode-folding-toggle-transactions ()
+  "Toggle hiding of all transactions in the current buffer."
+  (interactive)
+  (if (not (and (boundp 'hs-minor-mode) hs-minor-mode))
+      (message "Enable hs-minor-mode to use this functionality.")
+    (hs-life-goes-on
+     (save-excursion
+       (goto-char (point-min))
+       (while (re-search-forward "^[=~[:digit:]]" nil t)
```
bcc32, line 240: Can this reuse an already-defined regexp like `ledger-iterate-regexp`?

ledger-occur.el
drdv: > Is that intentional?

No

Probably defining as a user customization is better. I wasn't sure whether this PR is interesting to begin with, so I didn't polish such things. Furthermore, such definitions should be moved elsewhere but I would prefer for a person that knows the code well to make the final touches.

ledger-occur.el
drdv: Done

ledger-occur.el
drdv: I couldn't find a predefined regex that does the same thing (maybe I have missed something). Note that the code doesn't hide only transaction blocks but blocks like
```
= /Me:?/ & %/\\bSHARED\\b/
    (Me->You)                               0.5
```
Indeed, it would be better to define the regex in `ledger-regex.el` for consistency. A name like `ledger-block-start` seems a bit generic (any suggestions?).

drdv:
> Why define separate commands for toggling hide/show specifically for ledger-mode, when hideshow already defines such commands? Couldn't we just configure `hs-find-block-beginning-func` and the other corresponding functions for ledger-mode and then users could invoke the usual `hs-*` commands?
> 
> Just to confirm: if I'm understanding the code correctly, it is using hideshow's functionality to provide some overlay-related helpers and also things like the `'code` overlay type (to support isearch temporarily displaying the contents). But it doesn't really use `hs-minor-mode` except for the invisibility spec, right?

A great observation! 
Indeed the code doesn't use `hs-minor-mode` in the intended way. The reason is simple: I couldn't make it work with the standard mechanism (I remember trying but I guess I didn't put enough effort so I resorted to a custom implementation because there I have full control). All this is not ideal because, as you noted, I define separate commands -- which leads to some (minor) inconveniences in my user config and overall can be surprising to experienced people. 

Unfortunately, I don't have time right now to try again with the builtin `hs` mechanism.

bcc32:
Hmm, could we just use `outline-minor-mode` instead?  I think that would be better than trying to maintain some custom code.

In researching whether we could just use hs-minor-mode by configuring its regexps appropriately, I unfortunately ran into lots of performance issues.  hideshow uses `looking-back` in a critical section which causes folding the entire buffer to be quite slow (for my 68,000 line---3 megabyte---ledger file, it took about 1.5 seconds).

However, all we really need is a flat list, nothing with fancy hierarchy support like hideshow or origami.

So, if you just do:

```lisp
(setq-local outline-regexp "[^[:space:]]")
(outline-minor-mode)
```

then you can use `outline-minor-mode`'s commands to do pretty much what this PR provides, without any custom code.  (Plus, `outline-minor-mode` is popular and has lots of integrations such as with evil-mode).

The texinfo manual for ledger-mode mentions using outline-minor-mode in this way (but I think more for organization rather than individual xact-by-xact folding):

```
(eval-after-load 'ledger-mode
  (progn
    ;; org-cycle allows completion to work whereas outline-toggle-children does not
    (define-key ledger-mode-map (kbd "TAB") #'org-cycle)
    (add-hook 'ledger-mode-hook #'outline-minor-mode)
    (font-lock-add-keywords 'ledger-mode outline-font-lock-keywords)))
```

I think it's probably okay if we decide to provide a default `outline-regexp` in ledger-mode itself.  That way, users can still override it with file-local variables or `ledger-mode-hook`, but then out-of-the-box users of `outline-minor-mode` would get a useful value.

bcc32:
BTW `outline-minor-mode` also has very good performance, both compared to hideshow for this application, as well as compared to this PR (about 3x faster than the custom code using hideshow here).

drdv:
I learned something today ... thanks! I was not aware of `outline-minor-mode`. Indeed, for our envisioned folding problem, using it is as simple as setting `outline-regexp` (`"[^[:space:]]"` seems to work fine). 

I think that this PR can be closed.

drdv:
> In researching whether we could just use hs-minor-mode by configuring its regexps appropriately

Do you mind sharing your code?

bcc32:
> > In researching whether we could just use hs-minor-mode by configuring its regexps appropriately
> 
> Do you mind sharing your code?

Sure, it was just adding this in `ledger-mode.el` (after the `auto-mode-alist` definition):

```
(add-to-list 'hs-special-modes-alist `(ledger-mode ,ledger-iterate-regexp
                                                   ,ledger-iterate-regexp
                                                   ,ledger-comment-regex
                                                   ,(lambda (arg)
                                                      (dotimes (_ (- arg 1))
                                                        (ledger-navigate-next-xact))
                                                      (ledger-navigate-end-of-xact))
                                                   nil
                                                   ledger-navigate-beginning-of-xact))
```

bcc32:
Going to implement some defaults for outline-minor-mode in #449.  Will close this PR in favor of that.

Thank you for your contribution, and sorry it took so long to resolve!

PR #276, opened 2020-12-19
Make ledger-report-quit less opinionated about windows

jabranham:
ledger-report-quit's current behavior is to reset the window
configuration to whatever the window configuration was when the report
was originally called. This can be jarring when the report buffer was
generated, then the user went off doing other things (changing the
window configuration), then quits the report window.

The new behavior simply quits (and kills) the report buffer using
quit-windows-on, which works well with the display-buffer framework.


jabranham:
@purcell the tests seem to be broken on github actions

purcell:
Thx, I'm looking into the github action issue but cachix is having some issues so I'm currently being thwarted.

purcell:
CI is fixed now btw.

PR #278, opened 2021-01-01
Add ledger-report-native-highlighting-arguments

DamienCassou:
This removes one blocking point when displaying reports with `ledger-binary-path` pointing to hledger.


purcell:
Thanks!

PR #281, opened 2021-01-02
feat(xact): add default time when copying transaction

Fuco1:
Use the transaction's time as default time when asking when the copy should be
created.  This allows us to use the relative syntax of org-read-date to move the
transaction an <interval> forward or backward from the time of the transaction:

+1d (+ 1 day from TODAY)
++1d (+ 1 day from DEFAULT-TIME, i.e. date of the original transaction)

In other words, single +/- is relative against today and double ++/-- is
relative against the DEFAULT-TIME.

Note that since currently the DEFAULT-TIME is today, this feature is only
strictly adding functionality and should therefore be fully backward-compatible.


Fuco1:
For some reason the default time is shifted one day back, I think it might be some time-zone related issue with org read date.

Fuco1:
Oh, this actually changes the behaviour when user only hits RET upon the query.  It will copy to the original date instead of today.  I think that is a significant breaking change unfortunately :(

I still want to have some option to add past transactions easily, so I'll play with this a little bit more.

PR #294, opened 2021-03-16
ledger-add-transaction follows ledger-default-date-format

cacology:
## Problem:

`ledger-add-transaction` does not respect the value of `ledger-default-date-format` so for my ledger, it inserts dates as "2021/03/16" when I use "2021-03-16"

## Solution:

The command line argument `-y` passed to ledger with the value of that variable.  Viz. <https://www.ledger-cli.org/3.0/doc/ledger3.html#Output-customization>

Respectfully submitted,


cacology:
...I think I really miffed things up with getting the commits straight.  No ego here if you just want to throw this one away and do it right.  I think it's a useful patch.

cacology:
This is too messed up. Disregard and accept my apologies, please.

PR #296, opened 2021-03-17
Fix #295

### LINKED ISSUE #295: ledger-post-edit-amount produces multiple stack entries when amount has thousands separator
With the following example:

```
2021-01-01 Payee
  Expenses:Groceries    1,234.00 USD
  Liabilities:MasterCard
```

If the user presses `ledger-post-edit-amount` on the line with `1,234.00`, the calc stack now contains two entries, one for `1` and one for `234`.

Expected behavior: only one stack entry should be added, with value `1234.00` (or however calc's settings display that value).

Pull request on the way.

bcc32:
Fix #295 by removing thousands separators and converting decimal commas from the amount string before passing it to `calc-eval`.

### DIFF
--- a/ledger-commodities.el
+++ b/ledger-commodities.el
@@ -119,7 +119,7 @@ See `ledger-environment-alist' for DECIMAL-COMMA."
                       (assoc "decimal-comma" ledger-environment-alist))
                   (ledger-strip str "[.]")
                 (ledger-strip str ","))))
-    (while (string-match "," nstr)  ;if there is a comma now, it is a thousands separator
+    (while (string-match "," nstr)  ;if there is a comma now, it is a decimal point
       (setq nstr (replace-match "." nil nil nstr)))
     (string-to-number nstr)))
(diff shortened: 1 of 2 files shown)


purcell:
I don't think it's that simple. Ledger also supports European-style numbers in which commas denote decimals and periods are used for thousands separators.

bcc32:
I'm not sure I understand.

This PR uses `ledger-string-to-number`, which I assume is intended to handle the user's preference for European-style numbers.  Calc requires that input uses `.` for the decimal radix and `,` to separate stack elements, so the default `number-to-string` converts a float into the desired format.

purcell:
Ah, got it, should be fine then.

purcell:
I've fixed the underlying build issue (due to new byte compiler warnings in Emacs HEAD) so this is good to go - thanks!

bcc32:
Thanks, I appreciate the quick merge!

PR #298, opened 2021-03-22
Define ledger-iso-date-format before ledger-default-date-format

bcc32:
This fixes the following customization:

    (custom-set-variables
     '(ledger-default-date-format ledger-iso-date-format))

Otherwise, when `ledger-init` is loaded, Custom attempts to evaluate
`ledger-iso-date-format` immediately after `ledger-default-date-format` is
defined but before `ledger-iso-date-format` is defined, which leads to an
unbound variable error.


purcell:
Interesting edge case. Merging, thanks.

PR #303, opened 2021-04-27
Fix payee including trailing newline

### LINKED ISSUE #302: context payee includes trailing newline when transaction has comment
Given the following example transaction:

```
2016/08/12 KFC
 ; xact note
 a  $7
 b
```

Evaluating `(ledger-context-at-point)` on the first line of the transaction will return a context whose payee includes the trailing newline, namely: `"KFC\n"`.  This causes, among other things, `ledger-report-payee-format-specifier` to default to an incorrect value for the payee at point.

Pull request incoming.

bcc32:
Fix #302 by "moving" the newline from the payee regex to the comment regex.


purcell:
Yup, that seems reasonable, thanks.

PR #308, opened 2021-04-28
Fix #307: ledger-add-transaction inserts in the wrong place when xacts have effective dates 

### LINKED ISSUE #307: ledger-add-transaction inserts in the wrong place when xacts have effective dates
Given the following journal:

```
2013/05/02=2013/05/03 foo
    Expenses:Foo                            $10.00
    Assets:Bar

2013/05/03 foo
    Expenses:Foo                            $10.00
    Assets:Bar
```

Using `ledger-add-transaction` to insert a new transaction with date `2013/05/01` incorrectly places the new transaction between the two existing ones, rather than before the first one as expected.

bcc32:
This pull request fixes #307.  This branch itself includes some commits from #305 and #267, but I didn't know how to express the dependency in GitHub Pull Request terms.  In any case, those should probably be reviewed first (#267 is a refactor and also includes some bugfixes).

### DIFF
--- a/ledger-exec.el
+++ b/ledger-exec.el
@@ -89,6 +89,7 @@ otherwise the error output is displayed and an error is raised."
                    (apply #'call-process-region
                           (append (list (point-min) (point-max)
                                         ledger-binary-path nil (list outbuf errfile) nil "-f" "-")
+                                  (list "--date-format" ledger-default-date-format)
                                   args)))))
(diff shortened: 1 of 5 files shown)


purcell:
Thanks for all these. :+1:

bcc32:
Thanks for the fast merge!

PR #309, opened 2021-04-28
Remove unnecessary macroexpand calls in ledger-regex

bcc32:
One more for ya :)

### DIFF
--- a/ledger-regex.el
+++ b/ledger-regex.el
@@ -324,7 +324,7 @@
   "")
 
 (ledger-define-regexp full-amount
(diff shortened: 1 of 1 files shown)


purcell:
This seems to have failed the build, but I agree it should be possible to simplify this code a little.

bcc32:
Ah, it appears that arbitrary lisp in `regexp` is only supported from Emacs 27 onwards.  I'm keeping the removals that are still okay in emacs<27, although I'm also happy to just close the PR instead.

purcell:
> I'm also happy to just close the PR instead.

Nope, this is a good tidy-up, thanks!

PR #312, opened 2021-05-16
Fix ledger-narrow-on-reconcile

### LINKED ISSUE #311: ledger-narrow-on-reconcile has no effect
When `ledger-narrow-on-reconcile` is non-nil, the main buffer is supposed to be narrowed to show only transactions matching the account being reconciled.  This appears not to ever happen, and instead I get messages like the following and all transactions are visible in the main buffer:

```
No matches found for ’Expenses:Tax’
```

I think this bug was introduced a long time ago when `ledger-occur` changed so that it no longer has a `buffer` argument and instead uses the current buffer.  Will submit a PR to fix.

bcc32:
Fix #311 by calling ledger-occur in the main ledger buffer.  Also, we may as well regexp-quote the account name, since it may contain regexp metacharacters.

### DIFF
--- a/ledger-reconcile.el
+++ b/ledger-reconcile.el
@@ -554,10 +554,10 @@ moved and recentered.  If they aren't strange things happen."
       (add-hook 'after-save-hook 'ledger-reconcile-refresh-after-save nil t)
 
       ;; Narrow the ledger buffer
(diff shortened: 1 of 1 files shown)


purcell:
LGTM, thanks

PR #313, opened 2021-07-20
Ignore generated doc/ledger-mode.info

league:
Extraordinarily minor, but this is an attempt to be friendlier to the borg package manager for emacs. Borg uses git submodules, and if you configure it to find the `doc` directory like this:

```ini
[submodule "ledger-mode"]
	path = lib/ledger-mode
	url = git@github.com:ledger/ledger-mode.git
	info-path = doc
```

then it will build the `.info` file and configure the Emacs info path to find it, so you're reading it in-place in the submodule's working directory. (It doesn't get installed elsewhere.)

But then that `.info` file sitting there makes the submodule working directory appear "dirty", so whenever we're updating submodules it has to be worked around, or added to user's global ignore file. Thanks!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 *.elc
 *~
+/doc/ledger-mode.info


league:
@jdek No, closing. My work-around was to add `*.info` to global git ignore, then it covers other projects too. (If git is already tracking the `.info` in some project, then the ignore declaration is ignored.) Thanks.

PR #322, opened 2021-12-07
Replace reference to non-existing face variable

bram85:
This makes exports fail when Ledger snippets are embedded in an Org file.


purcell:
Maybe it was actually meant to be `ledger-font-default-directive-face`?

bram85:
> Maybe it was actually meant to be `ledger-font-default-directive-face`?

Yes, I revised my commit.

purcell:
Nice, thanks — merged.

PR #323, opened 2021-12-14
Update Actions versions

purcell:
Hoping to address build failures in PRs, which might also be caused by incompatiblities in the nix file in the ledger repo itself.


purcell:
Merging now #2082 is merged, this won't make `master` any more broken.

purcell:
Now seems fixed, thanks @akirak (https://github.com/ledger/ledger/pull/2085)

PR #326, opened 2022-01-22
[ledger-report] Delay report command expansion

### LINKED ISSUE #325: Editing a Report Command Edits the Expanded Form
When editing a command using `ledger-report-edit-report`, the user is presented with the `ledger-report-format-specifiers` already expanded.  This means if they save the edit, the format specifier values will then be hard coded.

I intend to open a PR to store the un-expanded report command in `ledger-report-cmd` and wait until `ledger-do-report` to actual expand the command.  I wanted to open this issue to document the problem.

kevinjfoley:
Delays the expansion of the report command until execution time.  This leaves
the command in `ledger-report-cmd' unexpanded, which prevents the expanded
version of the command from being edited / saved by the user.

Resolves #325

### DIFF
--- a/ledger-report.el
+++ b/ledger-report.el
@@ -471,7 +471,6 @@ Optionally EDIT the command."
     (when (or (null report-cmd) edit)
       (setq report-cmd (ledger-report-read-command report-cmd))
       (setq ledger-report-saved nil)) ;; this is a new report, or edited report
-    (setq report-cmd (ledger-report-expand-format-specifiers report-cmd))
     (set (make-local-variable 'ledger-report-cmd) report-cmd)
     (or (ledger-report-string-empty-p report-name)
         (ledger-report-name-exists report-name)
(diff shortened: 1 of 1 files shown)


enderw88:
Good idea

kevinjfoley:
> Good idea

Thanks!  Anything else needed?  I'm not sure what's wrong with the CI, I tried rebasing on the latest changes from master but it looks like it still isn't working.

bcc32:
I think this PR breaks a couple things.  For one thing, it causes the user to be re-prompted for any format specifiers that imply a prompt, such as `%(account)`, every time the report is run (e.g., if they modify and save the journal file).  Maybe the most robust way to address that issue would be to cache the expansion of each format specifier in some buffer-local alist or something, or maybe the expanded and unexpanded commands should both be saved so that the appropriate one can be used at each call site.

bcc32:
Superseded by #468

PR #332, opened 2022-05-16
Allow ledger-reconcile-insert-effective-date to be a function

bcc32:
Allow the user to customize ledger-reconcile-insert-effective-date by making it
a function that is called with the point at the posting to be cleared.  This
allows the function to decide whether to insert an effective date based on,
e.g., the payee or the account in question.

This is kept backwards-compatible with existing boolean values, which just
behave as a constant-returning function.

### DIFF
--- a/ledger-reconcile.el
+++ b/ledger-reconcile.el
@@ -127,8 +127,12 @@ no sorting, i.e. using ledger file order."
   :group 'ledger-reconcile)
 
 (defcustom ledger-reconcile-insert-effective-date nil
-  "If t, prompt for effective date when clearing transactions."
-  :type 'boolean
+  "If t, prompt for effective date when clearing transactions.
+
+If this is a function, it is called with no arguments with point
+at the posting to be cleared.  The return value is then used as
+described above."
+  :type '(choice boolean function)
   :group 'ledger-reconcile)
 
 (defcustom ledger-reconcile-finish-force-quit nil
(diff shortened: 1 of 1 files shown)


ledger-reconcile.el
```diff
@@ -215,6 +219,14 @@ And calculate the target-delta of the account being reconciled."
       (car where)
     (error "Function ledger-reconcile-get-buffer: Buffer not set")))
 
+(defun ledger-reconcile-insert-effective-date ()
+  "If `ledger-reconcile-insert-effective-date' is t, or returns
```
purcell, line 223: Would you mind making this comment conform to `checkdoc` please?

ledger-reconcile.el
```diff
@@ -229,9 +241,8 @@ And calculate the target-delta of the account being reconciled."
         (setq status (ledger-toggle-current (if ledger-reconcile-toggle-to-pending
                                                 'pending
                                               'cleared)))
-        (when ledger-reconcile-insert-effective-date
-          ;; Ask for effective date & insert it
-          (ledger-insert-effective-date)))
+        ;; Ask for effective date & insert it, if enabled
+        (ledger-reconcile-insert-effective-date))
```
purcell, line 245: But surely here this might sometimes be a bool, not a function?

purcell (changes requested):

ledger-reconcile.el
bcc32: This is the function defined above, not the defcustom (otherwise I'd have to `funcall` it).

ledger-reconcile.el
bcc32: Got it, done!

ledger-reconcile.el
purcell: Sorry, yes, ignore me. This seems fine as-is.

PR #333, opened 2022-05-17
Sort dates without leading 0 naturally

marcowahl:
E.g. sort 2022/5/7 as expected.  Previously one had to write 2022/05/07 to do  the sorting reliably.  Note the zeros!

The price for the fix of the sorting is the work to parse the date for each record.

### DIFF
--- a/ledger-sort.el
+++ b/ledger-sort.el
@@ -27,6 +27,7 @@
 ;;; Code:
 (require 'ledger-regex)
 (require 'ledger-navigate)
+(require 'ledger-xact)
 
 (defun ledger-sort-find-start ()
   "Find the beginning of a sort region."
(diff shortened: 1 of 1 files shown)


purcell:
Or you could substitute single-digits for zero-padded ones if parsing is costly?

marcowahl:
Yes that could be done.

The patch is thought to eliminate confusion when the user ledger-sorts a ledger with some dates without leading zeros.  The patch helps me and I can imagine that it can help others.

For me the parsing is not a time or performance issue.  Possibly it is for some people.

I think it's a correctness over speed question here.

bcc32:
I've gone ahead and rebased the PR, and benchmarked the performance with the following Lisp code in my personal ledger file (I have a ledger file with 10,000 xacts, total size 2.5MB):

```
(benchmark-run 10 (ledger-sort-buffer))
```

Before:

```
(15.29602 3 2.851102999999995)
```

After:
```
(16.228317 3 2.797007000000008)
```

So, it looks like a 6% slowdown in the overall sorting routine, which seems fine to me?

On a smaller file with just 3 xacts (and running `ledger-sort-buffer` 1000 times instead), I got before:
```
(0.7417469999999999 0 0.0)
```
and after:
```
(0.763527 0 0.0)
```

so only a 3% slowdown there.

@purcell, if you agree, I'll go ahead and rebase/squash this onto master.

purcell:
Yeah, that's totally fine IMO.

PR #334, opened 2022-05-23
Call ledger-reconcile-insert-effective-date in save-excursion

bcc32:
Thanks for merging #332, @purcell!  Just a quick follow-up because I forgot to wrap the funcall in a `save-excursion`.  I didn't judge `save-match-data` to be necessary here but I may well have missed something.


purcell:
Generally you wouldn't wrap with `save-match-data` explicitly unless the invoked function was known to intentionally set the match data.

PR #336, opened 2022-06-04
Shorten docstrings

### LINKED ISSUE #335: Docstrings need shortening for 28.1 byte compiler
Too-long docstrings now cause compilation issues, as of Emacs 28.1: we'll need to go through all the docstrings and shorten them. Noting this here as a collective to-do.

purcell:
Fixes #335

### DIFF
--- a/ledger-complete.el
+++ b/ledger-complete.el
@@ -156,7 +156,8 @@ Then one of the elements this function returns will be
 
 (defun ledger-accounts-list-in-buffer ()
   "Return a list of all known account names in the current buffer as strings.
-Considers both accounts listed in postings and those declared with \"account\" directives."
+Considers both accounts listed in postings and those declared
+with \"account\" directives."
   (let ((accounts (ledger-accounts-in-buffer)))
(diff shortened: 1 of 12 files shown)


purcell:
Ledger snapshot builds are still failing (as before), but otherwise this gets the CI working again, so will merge.

bcc32:
I think e8661a1 is broken, shouldn't it be something like the following?

```elisp
(defun ledger-exec-success-p (exit-code ledger-output-buffer)
  "Return non-nil if EXIT-CODE and LEDGER-OUTPUT-BUFFER indicate success."
  (and (zerop exit-code)
       (with-current-buffer ledger-output-buffer
         (goto-char (point-min))
         (not (and (> (buffer-size) 1) (looking-at (regexp-quote "While")))))))
```

Right now, it's returning `t` when `exit-code` is 1.

purcell:
@bcc32 Yes, oops, thanks — fix pushed as above.

purcell:
Super cool that the tests passed anyway... 😬

PR #340, opened 2022-06-14
Fix amount highlighting when tab is a separator

### LINKED ISSUE #338: Amount is not highlighted when separated from account name with a single tab (\t)
In the case of:

```ledger
2022-06-14 Decription
	Account1	10.00 PLN ; comment 1
	Account2	-10.00 PLN ; comment 2
```
(the separator is a single '\t')

the amounts and comments are not highlighted.

marek-g:
Fixes #338

### DIFF
--- a/ledger-regex.el
+++ b/ledger-regex.el
@@ -404,7 +404,7 @@
   (concat "^[[:blank:]]+"                 ; initial white space
           "\\(\\([*!]\\)?"                ; state and account 1, state 2
           "[[:blank:]]*\\(.*?\\)\\)?"     ; account 3
-          "\\(?:\t\\|[[:blank:]]\\{2,\\}" ; column separator
+          "\\(?:\\(?:\t\\|[[:blank:]]\\{2,\\}\\)" ; column separator
           "\\([^;\n]*?\\)"                ; amount 4
           "[[:blank:]]*\\(;.*\\)?\\)?$"   ; comment 5
           ))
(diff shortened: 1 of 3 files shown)


ledger-regex.el
```diff
@@ -404,7 +404,7 @@
   (concat "^[[:blank:]]+"                 ; initial white space
           "\\(\\([*!]\\)?"                ; state and account 1, state 2
           "[[:blank:]]*\\(.*?\\)\\)?"     ; account 3
-          "\\(?:\t\\|[[:blank:]]\\{2,\\}" ; column separator
+          "\\(?:\\(?:\t\\|[[:blank:]]\\{2,\\}\\)" ; column separator
```
purcell, line 407: It feels like the extra level of `?:` is redundant here, no?

marek-g: I think it is required. I'm not an regex expert but when I'm testing nested groups here: https://regex101.com/ with `(?:a(b))` pattern and `ab` string then `b` is captured into a group. I have to change pattern to `(?:a(?:b))` to not capture `b` into a group.

And having a new group will break the regex expression as in `ledger-fonts.el`  amount is expected to be group #4 not #5.

But if we don't put the `\t|[::blank::]{2,}` in a group at all, there is a bug that `\t` doesn't work as a separator between account and amount. You can easily test with any online regex tester.

marek-g: Maybe there is a better solution than surrounding that part of expression with a group? As I said, I'm not a regex expert.

marek-g: But if you are asking if the change is also required in the `ledger-regex.el` (and not only in `ledger-fonts.el`) - it may be the other way. I had to change only `ledger-regex.el` to make highlighting working in my case.

I did the other changes to `ledger-fonts.el` to be consistent. I thought it will be needed too as the intent is the same - to use tab or double space as a separator.

purcell: I think the `?:` are non-numbered matches, so they _shouldn't_ affect the capture group number. But the rest makes sense — I'll go ahead and merge now, thanks!

marek-g: Yes, `?:` won't affect the capture group number, that's why I used it.

Thanks Steve!

PR #341, opened 2022-06-14
Fix account completion with tab separated amount

### LINKED ISSUE #339: Account completion deletes rest of the line when the separator is a single tab (\t)
In the case of:

```ledger
2022-06-14 Decription
	Account1	10.00 PLN ; comment 1
	Account2	-10.00 PLN ; comment 2
```
(the separator is a single '\t')

if I change `Account2` to something else (for example to  `Account3`) using account completion I end up with:

```ledger
2022-06-14 Decription
	Account1	10.00 PLN ; comment 1
	Account3
```

(the `-10.00 PLN ; comment 2` part is destroyed).

marek-g:
Fixes #339

### DIFF
--- a/ledger-complete.el
+++ b/ledger-complete.el
@@ -301,7 +301,7 @@ Looks in `ledger-accounts-file' if set, otherwise the current buffer."
              (skip-chars-forward "([") ;; for virtual accounts
              (setq start (point)))
            (setq delete-suffix (save-excursion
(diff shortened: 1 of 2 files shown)


purcell:
Merged also, thanks again!

marek-g:
Great! Thank you Steve for quick merges! I love ledger-mode - it's so much better than ledger plugins for other editors :)

PR #349, opened 2023-01-15
ledger-reconcile-change-target: Display balance after changing

bcc32:
After entering an erroneous target amount, and then correcting it with `ledger-reconcile-change-target`, I wanted to see that the difference from target was now 0, but unfortunately the balance wasn't displayed.

This PR makes `ledger-reconcile-change-target` display the balance after setting the target.


purcell:
LGTM, thanks

PR #350, opened 2023-02-06
Use ledger-master-file for Flymake when available

bigodel:
Currently, multiple file setup ends up with incorrent linting due to `ledger-flymake` only considering the current file. In multiple file setup, where you would have your account definition separate from the actual accounting, this would lead to Flymake incorrectly linting accounts, commodities, etc as missing, when they're not. This simple change simply checks if the `(ledger-master-file)` is a file or not, and if so use it, otherwise default to using the current file.

We could make it so that the `(file-exists-p ...)` check is removed, but I don't know what is the desired behaviour in this case. Should we just assume that the user is providing the correct master file?

You could test out this feature by using the following files:

`foo.ledger`

```ledger
2023-01-01 Foo Bar
    Foo                                         -$10
    Bar

2023-01-01 Foo Baz
    Foo                                          $10
    Baz

; Local Variables:
; ledger-master-file: "main.ledger"
; End:
```

`main.ledger`

```ledger
account Foo
account Bar

commodity $
    default

include foo.ledger
```

You'll need to set `ledger-flymake-be-pedantic` to `t`. With the current change on `foo.ledger` you should be seeing a warning about the account "Baz" not being defined. If you change `ledger-master-file` to some non-existent file and restart Flymake, you should see warnings about "Foo" not being defined.

### DIFF
--- a/ledger-flymake.el
+++ b/ledger-flymake.el
@@ -91,40 +91,39 @@ Flymake calls this with REPORT-FN as needed."
           ;; be simply suspended.
           (when (eq 'exit (process-status proc))
             (unwind-protect
-                 ;; Only proceed if `proc' is the same as
-                 ;; `ledger--flymake-proc', which indicates that
-                 ;; `proc' is not an obsolete process.
-                 (if (with-current-buffer source (eq proc ledger--flymake-proc))
-                     (with-current-buffer (process-buffer proc)
-                       (goto-char (point-min))
-                       ;; Parse the output buffer for diagnostic's
-                       ;; messages and locations, collect them in a list
-                       ;; of objects, and call `report-fn'.
-                       (cl-loop
-                             while (search-forward-regexp
-                                    ;; This regex needs to match the whole error.  We
-                                    ;; also need a capture group for the error message
-                                    ;; (that's group 1 here) and the line number
-                                    ;; (group 2).
-                                    (rx line-start "While parsing file \"" (one-or-more (not whitespace)) " line " (group-n 2 (one-or-more num)) ":\n"
-                                        (zero-or-more line-start "While " (one-or-more not-newline) "\n" )
-                                        (minimal-match (zero-or-more line-start (zero-or-more not-newline) "\n"))
-                                        (group-n 1 "Error: " (one-or-more not-newline) "\n"))
-                                    nil t)
-                             for msg = (match-string 1)
-                             for (beg . end) = (flymake-diag-region
-                                                source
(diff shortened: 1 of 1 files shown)


ledger-flymake.el
```diff
@@ -66,8 +67,9 @@ Flymake calls this with REPORT-FN as needed."
     (kill-process ledger--flymake-proc))
   ;; Save the current buffer, the narrowing restriction, remove any
   ;; narrowing restriction.
-  (let ((source (current-buffer))
-        (file (buffer-file-name)))
+  (let* ((source (current-buffer))
+         (master (ledger-master-file))
+         (file (if (file-exists-p master) master (buffer-file-name))))
```
purcell, line 72: I wonder if it would be better to raise an error if `ledger-master-file` is set but the file is missing. Silently working around it is risky.

bigodel: it isn't really working around, but rather falling back to the previous behaviour. another solution would be to offer the user a y-or-n prompt

purcell: Yes, I understand it's a fallback. However, I think _if_ it makes sense to ever use `ledger-master-file` for flymake (which seeme likely to me), then if the user has set `ledger-master-file` but not created that file, it would be better IMO if their syntax checks fail loudly due to that missing file than to succeed silently with possibly-inconsistent results.

bigodel: so you think it would be better to error out instead of providing the user with a prompt? i'm ok with both, btw

purcell, line 72: Yeah, I'd just do this:

```suggestion
         (file (or (ledger-master-file) (buffer-file-name))))
```
I'll go ahead and apply this and we'll see if anyone complains. :)

purcell:
It's worth noting that in any case the flymake support is a bit hokey — it only works if the file is saved, whereas flymake could run at any time during editing. It would be possible to fix this to use temporary files instead (Flymake has support for this). However, in that case these changes wouldn't really work, because the master file would never refer to the temporary copy of the current file. It's also possible to put your accounts list in an `accounts-list.ledger` or similar, then add `include accounts-list.ledger` at the top of individual ledger files, which is what I do (in addition to having a master file).

bigodel:
> It's also possible to put your accounts list in an `accounts-list.ledger` or similar, then add `include accounts-list.ledger` at the top of individual ledger files, which is what I do (in addition to having a master file).

i also do that, but on top of it i have a file for each year, so a`2023.ledger` for example gets included in my main file.

PR #353, opened 2023-03-11
ledger-context: Allow cost element to appear in posting line

### LINKED ISSUE #352: Inferred report account includes amount when cost is specified
Given the following example transaction:

```
2016/08/12 KFC
  a  7 AAPL @ $100
  b
```

with point at the beginning of the `a` posting, invoke `ledger-report` and select `account` (which has `%(account)` in the command line).

The inferred account is incorrectly `a  7 AAPL @ $100` rather than `a`.

bcc32:
Fixes #352


purcell:
Will close and re-open to hopefully cause the PR to get rebased on top of `master`, where CI is now fixed.

bcc32:
Thanks for the merge :)

PR #359, opened 2023-05-06
Declare ledger-post-account-alignment* variables safe

pabl0:
Traditionally indentation related customizations are considered safe as file local variables.


purcell:
Thanks!

PR #361, opened 2023-06-05
ledger-split-commodity-string: Recognize amounts less than 1

bcc32:
Prior to this change, ledger-read-commodity-string returned strange things like `'(53 "0.")` for an input of `0.53` (with a default commodity of USD, it should have returned `'(53 "USD")`).

I decided to preserve a current property of the regexes, which is that leading zeros are still disallowed on non-zero numbers.


purcell:
LGTM, thanks!

PR #362, opened 2023-06-13
Add support for + and = prefixes in amount and ability to add whitespace after them

ipvych:
Right now postings with = and + prefixes in amount (i.e. `= 100 EUR` are not indented when pressing tab, this PR fixes that and adds ability to put spaces after these prefixes.

Indent function was changed because when we allow whitespace before prefix current indent function will just align amount including prefix meaning things like `=           100 EUR` will be left as is. I think it is better to trim this whitespace. If this breaks some expected workflow then I can make this part toggleable via config or remove it and keep it a part of my personal config.
I only tested this by checking that indent works properly and I don't know if this could affect other places

### DIFF
--- a/ledger-regex.el
+++ b/ledger-regex.el
@@ -344,7 +344,7 @@
   (note end-note))
 
 (defconst ledger-amount-regex
-  (concat "\\(  \\|\t\\| \t\\)[ \t]*[-+=]?"
+  (concat "\\(  \\|\t\\| \t\\)[ \t]*[-+=]? *"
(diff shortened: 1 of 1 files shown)


ipvych:
Whitespace trim breaks ability to have comment at end of line with two whitespaces. Removing trim but not sure if I should keep the ability to have whitespace after prefix. It can result in this weird indentation but unless someone does this intentionally it should not happen.

purcell:
I think this looks reasonable as-is, so merging - thanks!

PR #364, opened 2023-07-22
Push mark before copying a transaction

mssdvd:


purcell:
Looks reasonable to me, and I think I can see why it's helpful, but as a general suggestion for when you send pull requests, please briefly describe the *why* for the change, not just the *what*. Maintainers generally don't want changes exactly, they want relatable improvements or fixes. Merging now — thanks!

PR #367, opened 2023-08-10
Make `narrow-to-defun` work on ledger transactions/directives

vderyagin:
I find narrowing to a single transaction useful sometimes, this change makes it so that `narrow-to-defun` (bound to `C-x n d` by default) will do just that.


purcell:
Good idea, thanks!

PR #368, opened 2023-08-12
Added additional expansion formats

daudi:
Added expansion formats for amount1 and amount2, and date1 and date2. These can be used to find transactions with exact amounts or on exact dates, or in ranges. For example, to show transactions between two dates, in the ledger reports you can use:

%(binary) -f %(ledger-file) reg %(account) --display 'd>= [%(date1)] and d <= [%(date2)]'

### DIFF
--- a/ledger-report.el
+++ b/ledger-report.el
@@ -70,6 +70,10 @@ specifier."
     ("payee" . ledger-report-payee-format-specifier)
     ("account" . ledger-report-account-format-specifier)
     ("month" . ledger-report-month-format-specifier)
+    ("amount1" . ledger-report-amount1-format-specifier)
+    ("amount2" . ledger-report-amount2-format-specifier)
+    ("date1" . ledger-report-date1-format-specifier)
+    ("date2" . ledger-report-date2-format-specifier)
     ("tagname" . ledger-report-tagname-format-specifier)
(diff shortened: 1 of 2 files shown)


bcc32:
It seems like it would be nicer for the date format specifier to prompt for the date using `ledger-read-date`, instead of making the user just type a string.

I'm a little confused by the proposed "amount" format specifier---is it just supposed to be some number, or a commoditized amount, or something else?

Also, it seems a little weird to define `date1` and `date2` with the expectation that any reports that only use a single date should use `%(date1)`?  FWIW, you can just use the same format specifier twice in one report.  You'll get prompted with the same string each time, but otherwise it works okay, and I feel it would make the defaults a little cleaner.

daudi:
All good points - I'm very much as the lower end of the elisp learning curve. 
1) Yes, `ledger-read-date` is better
2) Yes, just an amount. In my report definition I have `%(binary) -f %(ledger-file) reg %(account) --display 'amount == %(amount1) GBP'` so the commodity is hard-coded there, but now I realise that is somewhat limiting. I'll have a go at making it possible to specify the commodity
3) My misunderstanding, yes, I see that it can just use one version of `date`

Thanks for your feedback.

Issue #53, opened 2017-10-14
highlight-regexp doesn't work in ledger-mode ?


simonmichael:
I just tried using highlight-regexp to highlight "TODO" in yellow, but it doesn't work unless I turn off font-lock-mode. That's a pity, and as far as I know unusual, that they won't coexist ?

purcell:
The "TODO" was placed in comments, presumably?

simonmichael:
It can also be in description or account names, eg.

jstamant:
This behavior isn't happening for me (6 years later).

I came across this issue while submitting a pull-request for fixing the comment highlighting for postings.

When I use `highlight-regexp` (with `font-lock-mode` of course) to highlight all occurences of `TODO`, it successfully highlights them all, regardless of where they are. If `TODO` is found _anywhere_ in the file, it gets highlighted over the fontification specified by ledger-mode. I've tried it in tags, in directives, in transaction lines, in posting lines, globally, etc.

Running Emacs version 29.1, and the latest ledger-mode version from MELPA (version 4.0.0?)

doolio:
If @simonmichael can confirm then I think we can close this issue.

simonmichael:
Thank you, yes I can't reproduce in emacs 30.0.50 with a recent ledger-mode.

PR #372, opened 2023-09-19
Changed comment highlighting to match the ledger file specification

jstamant:
I changed the regex for fontifying postings so that inline comments are only fontified when they meet the ledger specification (according to [manual section 4.7.1](https://ledger-cli.org/doc/ledger3.html#Transactions-and-Comments:~:text=ACCOUNT%20%20AMOUNT%20%20%5B%3B%20NOTE%5D)).

This change has no functional impact on ledger-mode's operation, but it will help users learn proper ledger file format, ensuring that they don't run into issues with `ledger` or `ledger-mode`. If users don't put a "long space" in front of posting comments, `C-c C-b <ledger-post-edit-amount>` destroys all (or part) or the comment and places point in the wrong location.

**Rationale:**
Comments that are on the same line as a posting are supposed to have a "long space" in front of them (one tab or 2+ whitespace characters). The current code doesn't make this distinction on postings, and will highlight everything after a semi-colon as a comment.

For 4 years, I thought that posting comments were anything after a semi-colon, but that was incorrect, and the font-lock highlighting reinforces that myth.

**Other observations:**
The current comment highlighting works as intended on transaction lines and for `comment` and `test` directive blocks.  I also did not find any directly-associated issues to reference, but there are a few references to how "broken" the comment highlighting is in issues #53 and #71.

----------------

Here's what the highlighting looks like **before** my change: (two comments here are _not_ supposed to be highlighted as comments)

screenshot of highlighting before the changes

Here's what the highlighting looks like **after** my change: (only valid comments are highlighted as comments)

screenshot of highlighting after the changes

--------------

Here's a some raw journal content for anyone wanting to try this for themselves:
```
;; Don't forget to use M-x font-lock-fontify-buffer to refresh the highlighting!!!
2022-02-05 Payee
    ; A transaction-level comment
    Account  $1.00  ; An inline comment (2 preceding spaces)
    Account  $1.00 ; NOT an inline comment (only 1 preceding space)
    Account  $1.00; NOT an inline comment (no preceding whitespace)
    Account  $1.00    ; An inline comment (one preceding \t tab)
    Account  ; An inline comment (2 preceding spaces)
    Account ; NOT an inline comment (only 1 preceding space)
    Account; NOT an inline comment (no preceding whitespace)
    Account	; An inline comment (one preceding \t tab)    
    Liabilities
```

### DIFF
--- a/ledger-regex.el
+++ b/ledger-regex.el
@@ -401,12 +401,14 @@
           ))
 
 (defconst ledger-posting-regex
-  (concat "^[[:blank:]]+"                 ; initial white space
-          "\\(\\([*!]\\)?"                ; state and account 1, state 2
-          "[[:blank:]]*\\(.*?\\)\\)?"     ; account 3
-          "\\(?:\\(?:\t\\|[[:blank:]]\\{2,\\}\\)" ; column separator
-          "\\([^;\n]*?\\)"                ; amount 4
-          "[[:blank:]]*\\(;.*\\)?\\)?$"   ; comment 5
+  (concat "^[[:blank:]]+" ; initial white space
+          ;; state and account, subexp 1
+          "\\(\\([*!]\\)?"            ; state,   subexp 2
+          "[[:blank:]]*\\(.*?\\)\\)?" ; account, subexp 3
+          "\\(?:\\(?:\t\\|[[:blank:]]\\{2,\\}\\)"
+          "\\([^;\n]*?\\)\\)?"        ; amount,  subexp 4
+          "\\(?:\\(?:\t\\|[[:blank:]]\\{2,\\}\\)"
+          "\\(;.*\\)\\)?$"            ; comment, subexp 5
           ))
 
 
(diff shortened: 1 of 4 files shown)


purcell:
Great idea — do you think you could convert your example into an extra test in `tests/fontify-test.el`?

jstamant:
Done! Not a problem - thanks for pointing me towards the test file.

All tests passed when I ran `M-x ert-run-tests-interactively` after `M-x eval-buffer`. Wasn't sure what the easiest way was to run the tests - I'm on Windows, at the moment, without a good development environment set up.

purcell:
Looks like there's a small offset error here: https://github.com/ledger/ledger-mode/actions/runs/6251068824/job/16971358168?pr=372#step:6:483

Hopefully not a platform-specific line-ending quirk!

jstamant:
Found the issue - I added one single space before a posting comment in the `demo-ledger` string, which caused ledger-reconcile/test-027 to fail. Updated that test's expected ending position to match!

purcell:
Nice, thanks — merged now.

PR #375, opened 2023-10-17
Call `ledger-next-amount` on successful `xact`

krydos:
Hi, 
Current `C-c C-a` behavior is when user is prompted for `xact` search term and there is a similar transaction found, that transaction appears in the file like so:

```
|2023-10-17 grocery
    expenses:food                              10 AUD
    assets:debit-card
```
where `|` in the beginning is the cursor position. 
It is often the case though that amount of money I spend on groceries is different. It is actually the case for almost all transactions. 

I think it makes more sense to place cursor next to the amount so users can immediately edit it if needed. 
```
2023-10-17 grocery
    expenses:food                              |10 AUD
    assets:debit-card
```
Like so. 

ledger-mode already has `ledger-next-amount` function so I just call it in case xact returns anything.

Let me know what do you think.

### DIFF
--- a/ledger-xact.el
+++ b/ledger-xact.el
@@ -216,15 +216,17 @@ date is requested via `ledger-read-date'."
         (when (looking-at "\n*\\'")
           (setq separator ""))))
     (if (> (length args) 1)
-        (save-excursion
-          (insert
-           (with-temp-buffer
-             (apply #'ledger-exec-ledger ledger-buf (current-buffer) "xact"
-                    (mapcar 'eval args))
-             (goto-char (point-min))
-             (ledger-post-align-postings (point-min) (point-max))
-             (buffer-string))
-           separator))
+        (progn
+          (save-excursion
+            (insert
+             (with-temp-buffer
+               (apply #'ledger-exec-ledger ledger-buf (current-buffer) "xact"
+                      (mapcar 'eval args))
+               (goto-char (point-min))
+               (ledger-post-align-postings (point-min) (point-max))
+               (buffer-string))
+             separator))
+          (ledger-next-amount))
       (progn
         (insert (car args) " ")
         (save-excursion (insert "\n" separator))))))


purcell:
Thanks for submitting this interesting change. I understand the intent, but it seems slightly ad-hoc imo. For example, it's expected that the date will vary from txn to txn, so the user is interactively prompted for a new date, and that's part of the signature for the command. If it were good for all users, we'd probably do the same for the amount. I'd worry that jumping to a specific part of the txn might be a behaviour very specific to an individual workflow.

Another way to do things would be for us to add an extension point so that you could jump to that point yourself, though you can probably already achieve that by advising the function.

krydos:
@purcell I'm not sure I fully understand the point about `date will vary from txn to txn`. It is true and it is already what's handled by xact and ledger-mode. User is required to choose a date so the date is always either what's entered by user or today (default). There is no reason to change it manually unless user mistakenly picked a wrong date, which I expect way more rare case than a variable amount.

There is also `ledger-copy-transaction-at-point` which is to me very similar to `ledger-add-transaction` and it uses `ledger-next-amount` function too. It also asks for a date and then just moves the cursor to amount field.

I believe you're right about advising the function. I didn't try this approach but I guess it would work as a local solution. 

@bcc32 - that's cool! I didn't know about it. Unfortunately it doesn't work in my case as I use hledger and custom xact script but I think I can update my xact script to make it work. Thanks! 

Do you guys think I am missing a use case here? I don't really understand what the reason of having a cursor in the beginning of the transaction. There is no action that can be done from this place. As far as I got there are few "actionable" places where cursor can be and it is either after the date so user can set a transaction status (*, !) or the amount.

bcc32:
You can set the transaction status by pressing C-c C-c rather than manually inserting `*`, for example.  Also, other folks may have automation that depends on point being at the beginning of the transaction (which seems like a pretty reasonable default), and it is easy to navigate to anywhere else in the transaction from there.  Making this change would break their workflows.

Meanwhile, the most common thing I personally need to do to a newly inserted transaction is not changing amounts, but accounts (e.g., because I bought something different from the store).  I don't think it's obvious that changing the amount of the first posting is the most important next step.

krydos:
@bcc32 good point about other people automations and accounts to change. 
It seems like there are quite a few places where cursor might be and you guys are probably right - it's not the best idea to just choose one of these places randomly (or based on my preference only) and put the cursor there.

I'll just advice the function then. 

Thanks!

PR #378, opened 2023-12-04
ledger-complete: Read flycheck-ledger-pedantic in addition to ledger-flymake-be-pedantic

bcc32:


purcell:
Hmm, it remains extremely hacky to be modifying the results returned by this function based on globals.

bcc32:
I pushed a new commit that proposes the following change:

The following functions now take an optional argument `include-postings`, which, when non-nil, requests that accounts in postings are included:

- ledger-accounts-in-buffer
- ledger-accounts-list-in-buffer
- ledger-accounts-list
- ledger-find-accounts-in-buffer
- ledger-accounts-tree

And now, `ledger-complete-at-point` simply passes in the argument based on the values of `ledger-flymake-be-pedantic` and `flycheck-ledger-pedantic` as before.

However, this change **is potentially breaking** since the optional argument's default may not coincide with the corresponding setting of `ledger-flymake-be-pedantic` in any given user's config.  Using required arguments would also not work since existing callsites would not pass the argument.

PR #379, opened 2023-12-04
Fix bug-reference-bug-regexp to conform to its contract

bcc32:
Starting in Emacs 28.1, bug-reference emits a warning if bug-reference-bug-regexp's first capture group does not contain its other capture groups (2-10).  The first capture group is used to specify which part of the buffer should be fontified as a link, and so it should include the bug number (capture group 2).


purcell (approved):

purcell:
thanks!

PR #381, opened 2023-12-04
ledger-accounts-in-buffer: Don't include account subdirectives

### LINKED ISSUE #380: Account subdirectives are included in account list for completion
If I have the following accounts file (`ledger-accounts-file`):

```
account Expenses:Food
    note Use for dining out, etc.

account Expenses:Groceries
    note Use for food I cook myself
```

Then auto-completing accounts in my journal file includes the `note` lines, which are not real account names.

This only happens when `ledger-flymake-be-pedantic` is nil.

bcc32:
Account subdirectives (like assert and note) might match the account name regex but should not be included in the list.  Instead of scanning each line in the buffer for such occurrences, only scan occurrences in transactions.

Fixes #380.

### DIFF
--- a/ledger-complete.el
+++ b/ledger-complete.el
@@ -146,12 +146,15 @@ Then one of the elements this function returns will be
           ;; with directives.  But the name is a little misleading.  Should we
           ;; make a ledger-mode-be-pedantic and use that instead?
           (bound-and-true-p ledger-flymake-be-pedantic)
-        (goto-char (point-min))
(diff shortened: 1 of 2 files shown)


purcell:
CI failure has a byte-compilation issue because `pos` is unused. Presumably iterating txns moves point each time before calling the function?

bcc32:
Thanks for catching that, yes.  I have updated the PR to properly ignore that argument.

PR #384, opened 2023-12-16
Refresh ledger-occur overlays after adding a new transaction while reconciling

### LINKED ISSUE #383: ledger-reconcile-add sometimes leaves the new xact invisible
When `ledger-narrow-on-reconcile` is non-nil, if the newly added xact is in an otherwise-invisible part of the buffer (e.g., a date range that only includes hidden xacts), it does not become visible until the user calls `ledger-occur-refresh`.

I think `ledger-reconcile-add` should automatically call this function.  (will submit a PR to fix this bug)

bcc32:
Refresh ledger-occur overlays after ledger-reconcile-add

Fix #383.

This PR also includes a couple of drive-by refactors.

### DIFF
--- a/ledger-reconcile.el
+++ b/ledger-reconcile.el
@@ -302,13 +302,15 @@ Return the number of uncleared xacts found."
         (recenter)
         (ledger-highlight-xact-under-point)))))
 
-(defun ledger-reconcile-add ()
-  "Use ledger xact to add a new transaction."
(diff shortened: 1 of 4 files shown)


bcc32:
Hi @purcell, any objection to merging this?

purcell (approved):
This all looks pretty reasonable to me fwiw

PR #386, opened 2023-12-30
Rework ledger-post-edit-amount

bcc32:
I originally just wanted to add a small docstring tweak and suppress the "Mark
set" message from this command, but this PR also includes some drive-by
refactors.

Also, add a hint to the Calc welcome message suggesting the user to press C-u y,
which will insert the top stack value into the ledger buffer and close calc.

### DIFF
--- a/ledger-post.el
+++ b/ledger-post.el
@@ -163,26 +163,25 @@ regular text."
 (defun ledger-post-edit-amount ()
   "Call `calc-mode' and push the amount in the posting to the top of stack."
   (interactive)
-  (goto-char (line-beginning-position))
+  (beginning-of-line)
   (when (re-search-forward ledger-post-line-regexp (line-end-position) t)
-    (goto-char (match-end ledger-regex-post-line-group-account)) ;; go to the and of the account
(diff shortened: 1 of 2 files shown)


ledger-post.el
```diff
@@ -161,28 +161,33 @@ regular text."
...
+          (calc-eval (number-to-string (ledger-string-to-number val-string)) 'push))
+      ;; make sure there are two spaces after the account name and go to calc
+      (if (search-backward "  " (- (point) 3) t)
+          (end-of-line)
+        (insert "  "))
+      (push-mark (point) 'nomsg)
+      (calc))
+
+    ;; Preserve calc's own welcome message, if any, and append our own.
+    (message (substitute-command-keys "%sPress \\[universal-argument] \\[calc-copy-to-buffer] to use the top of stack as the new amount.")
```
purcell, line 187: Here, since the substituted keys for calc-copy-to-buffer could theoretically contain `%`, I'd recommend using "%s%s" as the format string, then providing the two messages as params.

bcc32: Good point, done.

PR #387, opened 2023-12-31
Make ledger-mode-clean-buffer remember point within transaction

DamienCassou:
* ledger-mode.el (ledger-mode-clean-buffer): Improve to make it
remember the point within the transaction. Before, the point was
always moved at the beginning of the current transaction, making the
function inconvenient for use within `before-save-hook'.


purcell:
Makes sense, thanks

PR #389, opened 2024-01-07
reconcile: Optionally display balance in header line

bcc32:
This commit adds a new customization,
`ledger-reconcile-display-balance-in-header`.  If non-nil, this causes the
reconcile buffer's header line to display the cleared+pending balance, as well
as the difference from target (if any), continuously while reconciling.  I have
found this far more convenient than searching through `*Messages*` for the last
balance message (although I learned that `ledger-display-balance` has a key
binding while implementing this, I still think this feature may be useful).

### DIFF
--- a/ledger-reconcile.el
+++ b/ledger-reconcile.el
@@ -150,6 +150,12 @@ described above."
   :group 'ledger-reconcile)
 
 (defvar-local ledger-reconcile-last-balance-message nil)
+(defvar-local ledger-reconcile-last-balance-equals-target nil)
+
+(defface ledger-reconcile-last-balance-equals-target-face
+  '((t :inherit header-line :foreground "green3"))
+  "Face used for header line when cleared-or-pending balance equals the target."
+  :group 'ledger-reconcile)
 
 ;; s-functions below are copied from Magnars' s.el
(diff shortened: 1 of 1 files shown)


ledger-reconcile.el
```diff
@@ -580,7 +599,15 @@ moved and recentered.  If they aren't strange things happen."
       (if ledger-narrow-on-reconcile
           (ledger-occur (regexp-quote account)))
 
+      (setq ledger-reconcile-last-balance-message nil)
+      (setq ledger-reconcile-last-balance-zero nil)
+
       (with-current-buffer rbuf
+        (when ledger-reconcile-display-balance-in-header
```
purcell, line 606: If I disabled this var, the pre-existing header-line-format would be left behind. Might it make sense to provide a little minor mode for this rather than a magic toggle var?

bcc32: Good point.  I have made it into a minor mode, and also took the liberty of binding it to the `B` key in the reconcile buffer (but I can take it out if it's objectionable).

purcell (approved):

bcc32:
Thanks for review!  I took the liberty of adding a NEWS entry for this one, too.

PR #396, opened 2024-01-17
Fix ledger auto report width

parkouss:
By using the `window-max-chars-per-line` instead of `window-width`, which seems exactly what we want here according to the doc:

Return the number of characters that can be displayed on one line.

This function automatically take care of the glyph column.


purcell (approved):

parkouss:
Thanks @purcell, I'll close the issue 395 that I opened to track the issue. Thanks for all the work on ledger-mode, this is fantastic.

PR #399, opened 2024-02-24
New 'ledger-indent-region' function

### LINKED ISSUE #197: Can't indent regions
Selecting the following region and pressing TAB which is bound to `indent-for-tab-command` does not indent the transaction properly.

```
2019/08/13 Test
this  8000 GBP
out
```

Going through each line and pressing TAB will indent it correctly.

jdek:
* ledger-mode.el (ledger-mode): Set indent-region-function.
* ledger-post.el (ledger-indent-region): New function that indents the region by calling ledger-indent-line for each line.

Closes #197

### DIFF
--- a/ledger-mode.el
+++ b/ledger-mode.el
@@ -331,7 +331,7 @@ With a prefix argument, remove the effective date."
   (ledger-init-load-init-file)
   (setq-local comment-start ";")
   (setq-local indent-line-function #'ledger-indent-line)
-  (setq-local indent-region-function 'ledger-post-align-postings)
+  (setq-local indent-region-function #'ledger-indent-region)
   (setq-local beginning-of-defun-function #'ledger-navigate-beginning-of-xact)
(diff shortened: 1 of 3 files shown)


purcell:
Probably @enderw88 would be better placed to comment on whether this is a desirable fix — I'm not sure about the history of indentation in ledger-mode

purcell:
As far as I can tell, the default `indent-region-function` – `indent-region-line-by-line` – calls `indent-line-function` on each line. So what's the difference here, assuming that `indent-line-function` is set to `ledger-indent-line`?

purcell:
ie. what you've implemented feels the same as simply not setting `indent-region-function` in the first place.

purcell:
> ledger-post-align-postings is broken somehow I guess?

Hmmm... do we have any tests for that, perhaps? You could consider pushing a failing test to show what you'd expect.

PR #404, opened 2024-03-17
Allow `ledger-insert-effective-date` to work with regions.

ChickenProp:
It updates or removes the effective date of all transactions that start in the region (technically, that start on a line in the region), without editing any postings.

With my credit card, I use the statement date as the effective date. That means I have a long block of transactions that all have the same effective date. This makes it easy to set them all at once.

### DIFF
--- a/ledger-mode.el
+++ b/ledger-mode.el
@@ -176,34 +176,75 @@ the balance into that."
                       nil 'noerr)
                  (replace-match ""))))))))
 
-(defun ledger-insert-effective-date (&optional date)
+(defun ledger-insert-effective-date (&optional start end date)
   "Insert effective date `DATE' to the transaction or posting.
 
-If `DATE' is nil, prompt the user a date.
+If `DATE' is nil, prompt the user for a date.
 
 Replace the current effective date if there's one in the same
 line.
 
-With a prefix argument, remove the effective date."
-  (interactive)
-  (if (and (listp current-prefix-arg)
-           (= 4 (prefix-numeric-value current-prefix-arg)))
-      (ledger-remove-effective-date)
-    (let* ((context (car (ledger-context-at-point)))
-           (date-string (or date (ledger-read-date "Effective date: "))))
-      (save-restriction
-        (narrow-to-region (line-beginning-position) (line-end-position))
-        (cond
-         ((eq 'xact context)
-          (beginning-of-line)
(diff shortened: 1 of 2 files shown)


ledger-mode.el
```diff
@@ -176,34 +176,75 @@ the balance into that."
                       nil 'noerr)
                  (replace-match ""))))))))
 
-(defun ledger-insert-effective-date (&optional date)
+(defun ledger-insert-effective-date (&optional start end date)
```
bcc32, line 179: Adding optional arguments _before_ other optional arguments is a breaking change to the Lisp interface; please add these at the end of the arguments list, instead.

ledger-mode.el
```diff
@@ -176,34 +176,75 @@ the balance into that."
...
+            (re-search-forward ledger-iso-date-regexp)
+            (when (= (char-after) ?=)
+              (ledger-remove-effective-date))
+            (insert "=" date-string))
+           ((eq 'acct-transaction context)
+            (end-of-line)
+            (ledger-remove-effective-date)
+            (insert "  ; [=" date-string "]"))))))))
+
+(defun ledger-insert-effective-date-region (start end &optional date)
```
bcc32, line 216: There's a lot of duplicated code between this function and `ledger-insert-effective-date`.  Can they be combined?  I would think that just looping over the lines in `start..end` and calling `ledger-insert-effective-date` (with the correct `current-prefix-arg`) should be sufficient.

bcc32 (changes requested):
Hi, thanks for your contribution!

In addition to the inline comments I have left, can you also please rebase your branch and add some tests for the new functionality?  I have just pushed some baseline tests for the existing `ledger-insert-effective-date` to the master branch.

ChickenProp:
Tests added. They're not as thorough as I'd like: in particular, I don't know how to test prompting behavior, and when refactoring I briefly introduced a bug that my tests didn't catch. (If I called `(ledger-insert-effective-date date)` instead of `(... date-string)`, then interactive use would ask for a different date for every xact in the region, plus one up front that would be ignored. Tests don't catch it because they never get dates from prompts.)

ledger-mode.el
ChickenProp: I managed to reduce it a bit, but there's still a decent amount of duplication.

PR #405, opened 2024-03-18
Refactor definitions and use of buffer-local variables

bcc32:
This PR cleans up some use of buffer-local variables in ledger checks, reports,
and reconcile buffers.  These variables are now properly namespaced and
designated as permanent buffer-locals, so they can be set straightforwardly with
setq.

I added docstrings only for the ledger-reconcile buffer-locals, because at least
one of them is used in a test.

While I consider this code mostly pretty straightforward, it does have the
potential to break somebody's personal config if they are referring to any of
these variables by the (un-namespaced) original name.  @purcell, thoughts?

### DIFF
--- a/ledger-check.el
+++ b/ledger-check.el
@@ -32,7 +32,7 @@
 
 
 (defvar ledger-check-buffer-name "*Ledger Check*")
-(defvar ledger-original-window-cfg nil)
+(defvar-local ledger-check--original-window-configuration nil)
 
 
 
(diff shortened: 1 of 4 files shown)


purcell:
> it does have the
> potential to break somebody's personal config if they are referring to any of
> these variables by the (un-namespaced) original name. @purcell, thoughts?

I don't think that's a concern for the particular variables you've renamed.

PR #406, opened 2024-03-18
ledger-report: Pass ledger-default-date-format to ledger command

### LINKED ISSUE #289: ledger-default-date-format setting not respected in reports
I've set this to match `ledger-iso-date-format`. The docstring `ledger-default-date-format` suggests the set format will be used by ledger throughout.  I took this to also include reports. It does not. At a minimum I would like the report dates to include the full year e.g. 2021 vs 21.

Perhaps if/when [ledger/issues/1853](https://github.com/ledger/ledger/issues/1853) is implemented this would be respected?

Thanks for your time.

bcc32:
Fix #289


bcc32:
I agree, and it does work if you just put it in .ledgerrc instead, but this custom variable already exists, and I feel we may as well make it do what it is documented to do.  In addition, it is at least useful to have such a variable to override the behavior in certain contexts.  However, I agree that the variable needs some thinking about.  It currently doesn't distinguish between output date format and input date format, which ledger-cli does, and in fact ledger-mode reads `--input-date-format` from ledgerrc.  It's all a bit messy, IMO.

bcc32:
I guess I was a little unclear in my previous comment---`ledger-format-date` is used when ledger-mode wants to insert a date into a ledger file, so defaulting to the user's preferred input format makes sense (i.e., if they specified one in ledgerrc, ledger-mode should automatically insert dates that way when editing ledger files).  The only place it is used for output is in reconciliation buffers (see below) and also in ledger-schedule, which I am unfamiliar with.

`--input-date-format` from ledgerrc is otherwise not passed to `--date-format`.  It is currently the case, I believe, that (with the exception of the bug this PR is designed to address), `--date-format <ledger-default-date-format>` is always passed to the ledger cli, which overrides ledgerrc's specification of `--date-format` if the latter is present.

For the reconciliation buffers, it looks like they do not actually use the `ledger-reconcile-default-date-format` setting at all right now.  I think we should fix that bug (opened #408 to track)

PR #407, opened 2024-03-19
Fix #54 and refactor ledger-occur overlay logic

### LINKED ISSUE #54: ledger-occur (C-c C-f) uses different face for last match ?
When I do C-c C-f PAT, all of the matched transactions are shown in ledger-occur-xact-face - except for the last, which remains in the normal face for some reason.

bcc32:
Fix #54.

This PR adds some regression tests for ledger-occur, as well as refactoring the
overlay creation loop to be a bit easier to understand.

Another option I considered was to use a single overlay for the whole buffer to
make most of it invisible, and then only create overlays to make the match areas
visible.  However, with ledger-occur-xact-face, it highlights the newlines in
between the xacts, which is a little weird, so I didn't end up making this
change.

### DIFF
--- a/ledger-occur.el
+++ b/ledger-occur.el
@@ -155,7 +155,9 @@ Argument OVL-BOUNDS contains bounds for the transactions to be left visible."
       (nreverse lines))))
 
 (defun ledger-occur-compress-matches (buffer-matches)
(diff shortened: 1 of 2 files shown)


purcell (approved):
Nice!

PR #410, opened 2024-03-22
Use correct regexp when scanning for account names

purcell:
This is a minor fix to a previous change from @bcc32 — would you mind checking it over please?

The previous loop searched for directives, and the intention of this second loop is to search only for account names used in transactions, so we don't need to use the combined regexp here after all.


bcc32 (approved):
Yep, LGTM

PR #392, opened 2024-01-09
Add user option for completion collection for payees

josephmturner:
This user option can be used to offer completion over a list of payees from a file:

``` elisp
(defvar-local my/ledger-payees-collection nil
  ;; TODO: If `project.el' adds support for project-local variables,
  ;; use that instead of `defvar-local'.
  "Memoized payees collection returned by `my/ledger-payees-collection'.")

(defun my/ledger-payees-collection ()
  "Return memoized list of payees contained in `my/payees-file'.
`my/payees-file' can be locally bound to a filepath string to a
file containing a newline-delimited list of payees, where the
filepath is relative to the current project root.  With nil or
nonexistent `my/payees-file', return `ledger-payees-in-buffer.'"
  (let ((payees-file
         (and (bound-and-true-p my/payees-file)
              (expand-file-name my/payees-file (project-root (project-current t))))))
    (cond ((not (and payees-file (file-exists-p payees-file)))
           #'ledger-payees-in-buffer)
          ((or (not my/ledger-payees-collection)
               (file-has-changed-p payees-file
                                   'my/ledger-payees-collection))
           (setf my/ledger-payees-collection
                 (with-current-buffer (find-file-noselect payees-file)
                   (split-string (buffer-string) "\n"))))
          (my/ledger-payees-collection))))

(setopt ledger-payees-collection #'my/ledger-payees-collection)
```

  To use this snippet, create a file containing a list of payees inside your project and set its relative filepath as a dir-local variable:

``` elisp
((ledger-mode . ((my/payees-file . "export/all-payees.txt"))))
```

PR #411, opened 2024-03-22
Add ledger-payees-file for completion, and support payee directives

purcell:
See #392


josephmturner:
Thanks!  #413 augments this PR.

PR #417, opened 2024-03-26
Fix #416 by allowing subtraction of zero-amount mismatched commodities

### LINKED ISSUE #416: ledger-reconcile raises an error if current balance is zero
When there are uncleared transactions for an account, but the cleared+pending
balance is zero, ledger-reconcile raises an error like so:

```
ledger-subtract-commodity: Can’t subtract different commodities (0 nil) from (0 "USD")
```

Arguably, subtracting zero from a commoditized amount (even if the commodities
don't match), should be fine.  In this case, it's just because ledger doesn't
print a commodity at all if the amount is zero.

bcc32:
This PR changes `ledger-{add,subtract}-commodity` to allow the second argument's
commodity string not to match the first argument's, as long as the second amount
is zero.  Since subtracting a zero amount is a no-op anyway, this should be fine
in terms of correctness, and alleviates some spurious errors.

Fix #416

### DIFF
--- a/ledger-commodities.el
+++ b/ledger-commodities.el
@@ -94,16 +94,26 @@ Returns a list with (value commodity)."
           (split-string str "[\n\r]")))
 
 (defun ledger-subtract-commodity (c1 c2)
-  "Subtract C2 from C1, ensuring their commodities match."
-  (if (string= (cadr c1) (cadr c2))
-      (list (-(car c1) (car c2)) (cadr c1))
-    (error "Can't subtract different commodities %S from %S" c2 c1)))
+  "Subtract C2 from C1, ensuring their commodities match.
+
+As an exception, if the quantity of C2 is zero, C1 is returned
+directly."
+  (cond
+   ((zerop (car c2)) c1)
+   ((string= (cadr c1) (cadr c2))
+    (list (- (car c1) (car c2)) (cadr c1)))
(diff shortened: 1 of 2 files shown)


purcell:
This feels like an indirect and unexpected fix to functions not directly related to ledger-reconcile. It's still a logical error to subtract/add mismatching commoditised values, whatever the numbers. I suspect there may be a better, more local, fix for #416.

bcc32:
I did think of that, but I think this change is still "justified", in that:

1. These functions are only used by reconcile (in fact, `ledger-add-commodity` is not used at all)
2. ledger itself makes this simplification:

```
$ ledger eval '0 USD'
0

$ ledger eval '0 USD - 0'
0

$ ledger eval '0 USD - 0 GBP'
0
```

FWIW, the commodity representation in ledger-mode is not quite accurate anyway; ledger itself can represent multiple commodities in the same amount just fine, although the printing and reading syntax then becomes more complex:

```
$ ledger eval '1 USD - 1 GBP'
-1 GBP
1 USD
```

(so I suppose it is reasonable not to support it directly in ledger-mode)

If you think this change should be more localized to reconciliation, I can make that change too, although since it's ledger itself that's printing `0` instead of `0 USD` (or whatever commodity), it feels a bit more hacky to me than just changing the subtraction code.

purcell:
That makes sense. Happy for this change to be merged as-is, then.

PR #421, opened 2024-04-10
ledger-xact: Add ledger-xact-fill to insert missing amount

DamienCassou:
If the changes in this PR are desired, I will write unit tests.

### DIFF
--- a/ledger-post.el
+++ b/ledger-post.el
@@ -200,7 +200,8 @@ the postings in the transaction.
 
 MISSING-POSITIONS is a list of positions in the buffer where the
 transaction do not have an amount specified (such postings do not
-contribute to TOTAL).
+contribute to TOTAL).  Specifically, the positions are at the end
+of the account name on such posting lines.
 
 Error if the commodities do not match."
   (save-excursion
--- a/ledger-post.el
+++ b/ledger-post.el
@@ -240,6 +240,9 @@ might want to insert it anyway."
          (goto-char missing-pos)
          (insert "  " (ledger-commodity-to-string missing-amount))
          (just-one-space 0)
+         ;; if point is not at the end of the line after the amount, it means
+         ;; there is a trailing comment that should be separated from the
+         ;; amount with a long space:
          (unless (equal (point) (line-end-position))
            (just-one-space 2))
          (ledger-post-align-xact (point))))
--- a/ledger-post.el
+++ b/ledger-post.el
@@ -229,17 +229,18 @@ might want to insert it anyway."
   (pcase-let* ((`(,total . ,missing-positions) (ledger-post-xact-total))
                (missing-amount (ledger-negate-commodity total))
                (amounts-balance (< (abs (car missing-amount)) 0.0001)))
-    (pcase (length missing-positions)
-      (0 (unless amounts-balance
-           (user-error "Postings do not balance, but no posting to fill")))
-      (1 (if amounts-balance
-             (user-error "Missing amount but amounts balance already")
-           (goto-char (car missing-positions))
-           (insert "  " (ledger-commodity-to-string missing-amount))
-           (just-one-space 0)
-           (unless (equal (point) (line-end-position))
-             (just-one-space 2))
-           (ledger-post-align-xact (point))))
+    (pcase missing-positions
+      ('() (unless amounts-balance
+             (user-error "Postings do not balance, but no posting to fill")))
+      (`(,missing-pos)
+       (if amounts-balance
+           (user-error "Missing amount but amounts balance already")
+         (goto-char missing-pos)
+         (insert "  " (ledger-commodity-to-string missing-amount))
+         (just-one-space 0)
+         (unless (equal (point) (line-end-position))
+           (just-one-space 2))
+         (ledger-post-align-xact (point))))
       (_ (user-error "More than one posting with missing amount")))))
 
 (provide 'ledger-post)
--- a/ledger-post.el
+++ b/ledger-post.el
@@ -193,14 +193,14 @@ the amount and return to ledger."
       (calc))))
 
 (defun ledger-post-xact-total ()
-  "Return (TOTAL . MISSING-COUNT) for the transaction at point.
+  "Return (TOTAL . MISSING-POSITIONS) for the transaction at point.
 
 TOTAL is a commoditized amount representing the total amount of
 the postings in the transaction.
 
-MISSING-COUNT is the number of postings in the transaction that
-do not have an amount specified (such postings do not contribute
-to TOTAL).
+MISSING-POSITIONS is a list of positions in the buffer where the
+transaction do not have an amount specified (such postings do not
+contribute to TOTAL).
 
 Error if the commodities do not match."
   (save-excursion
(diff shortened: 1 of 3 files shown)


bcc32:
I think this command should consider whether the amounts in the postings are actually for the same commodity, e.g., if you had a transaction like:

```
2024-04-01 Hotel
    Expenses:Lodging    200 EUR
    Expenses:VAT
    Liabilities:Mastercard  -225 USD
```

It would not be correct to insert "25 EUR" or "25 USD" in the missing spot.  (Also, it seems like this implementation does not insert a commodity symbol at all, is that intentional?)

DamienCassou:
@bcc32 thank you for your quick feedback, I appreciate it. I think I have fixed my implementation to taken commodities into account (I never write any so I forgot about this use case, sorry).

DamienCassou:
I changed the code to avoid a circular dependency between ledger-xact.el and ledger-post.el

bcc32:
Hi @DamienCassou, I pushed a couple more commits to the branch.  I think this fixes the commodity issue I was thinking of.

I'm a little unsure of what you think the correct behavior should be in this case:

```
2024-04-01 Hotel
    Expenses:Lodging                             225 USD
    Expenses:VAT
    Liabilities:Mastercard                      -225 USD
```

Currently the code will do nothing, but maybe it should insert an explicit zero amount?  I tried to preserve as much of behavior in your original version as possible

DamienCassou:
I'm so confused. You reviewed an old version of my work because I committed my changes to the wrong branch :-(. I'm sorry for your time. 

The version I wanted you to review had support for commodity and avoided a circular dependency between ledger-post and ledger-xact. I think your work is better anyway so I will ignore mine. Here are the 2 functions I wrote just in case they prove useful in the future:

```emacs
(defun ledger-xact-amounts ()
  "Return the amounts of the xact containing point or nil.

Each amount is of the form (VALUE COMMODITY), VALUE being a
number and COMMODITY a string."
  (save-excursion
    (cl-destructuring-bind (begin end) (ledger-navigate-find-xact-extents (point))
      (goto-char begin)
      (forward-line 1) ;; ignore the first line
      (let ((amounts))
        (while (< (point) end)
          (when-let* ((context (ledger-context-at-point))
                      ((eq (ledger-context-line-type context) 'acct-transaction))
                      (amount (ledger-split-commodity-string
                               (or
                                (ledger-context-field-value context 'commoditized-amount)
                                (ledger-context-field-value context 'amount)))))
            (setq amounts (cons amount amounts)))
          (forward-line 1))
        (nreverse amounts)))))

(defun ledger-xact-fill ()
  "Find a posting with no amount and insert it.

Even if ledger allows for one missing amount per transaction, you
might want to insert it anyway."
  (interactive)
  (cl-destructuring-bind (begin end) (ledger-navigate-find-xact-extents (point))
    (when-let* ((amounts (ledger-xact-amounts))
                (amounts-sum (cl-reduce #'ledger-add-commodity amounts))
                (missing-amount (ledger-subtract-commodity
                                 (list 0 (cadr amounts-sum))
                                 amounts-sum))
                (end-of-line-regex (rx (* (any space)) line-end)))
      (when (> (abs (car missing-amount)) 0.0001)
        (goto-char begin)
        (while (and (< (point) end)
                    (ledger-next-account end)
                    (match-end 0)
                    (progn
                      (goto-char (match-end 0))
                      (not (looking-at-p end-of-line-regex)))))
        (when (match-end 0)
          (insert "  " (ledger-commodity-to-string missing-amount))
          (ledger-post-align-dwim))))))
```

Please give me some more time to read your work, answer your question and maybe suggest some changes.

DamienCassou:
@bcc32: I added some tests and strengthen yours by checking the content of the error message as well.

Regarding the case where an amount is missing but the transaction already balance, I decided to change the code to throw an exception and added a corresponding test case. My opinion is that the user should be aware there is something fishy instead of silently hiding the issue. If you disagree, I can implement what you have in mind instead.

If you agree with everything, feel free to squash all commits together and put your name there: there is not much code from me anymore anyway :-).

I'm sorry again for asking you to review the wrong version of my work. This definitively lead to more work for you.

DamienCassou:
I found a bug, working on it.

DamienCassou:
@bcc32 bug fixed. I had to refactor ledger-post-fill to avoid the duplication of the code searching for the missing amount. What do you think?

DamienCassou:
I just removed the definition of 2 unused variables.

ledger-post.el
```diff
@@ -188,6 +192,56 @@ the amount and return to ledger."
...
+
+Error if the commodities do not match."
+  (save-excursion
+    (pcase-let ((`(,begin ,end) (ledger-navigate-find-xact-extents (point))))
+      (goto-char begin)
+      (cl-loop
+       while (re-search-forward ledger-post-line-regexp end t)
+       for amount-string = (when-let ((amount-string (match-string ledger-regex-post-line-group-amount)))
+                             (unless (string-empty-p (string-trim amount-string))
+                               amount-string))
```
bcc32, line 213: I'm surprised there can be a case where `(string-empty-p (string-trim amount-string))` is true.  The definition of `ledger-post-line-regexp` seems to already account for that by consuming the long space before and after the amount.  Maybe there is a possibility I am missing here?

ledger-post.el
```diff
@@ -188,6 +192,56 @@ the amount and return to ledger."
...
+    (pcase (length missing-positions)
+      (0 (unless amounts-balance
+           (user-error "Postings do not balance, but no posting to fill")))
+      (1 (if amounts-balance
+             (user-error "Missing amount but amounts balance already")
+           (goto-char (car missing-positions))
+           (insert "  " (ledger-commodity-to-string missing-amount))
+           (just-one-space 0)
+           (unless (equal (point) (line-end-position))
+             (just-one-space 2))
```
bcc32, line 241: This seems a little mysterious and possibly overcomplicated.  Perhaps we should just rely on the user to have trailing spaces removed by something like `whitespace-cleanup` on save, rather than doing an ad-hoc version of it here?

bcc32:
I like the new approach overall, and thanks for the new tests.  Added a couple of minor improvements to the branch directly, but I had a couple questions about some details (in the review interface).

ledger-post.el
DamienCassou:
```emacs
(let ((ledger "    Assets:Bar        "))
  (string-match ledger-post-line-regexp ledger)
  (match-string-no-properties
    ledger-regex-post-line-group-amount
    ledger))
```

returns `"      "` containing 6 spaces: this is the number of trailing spaces in `ledger` (8) minus the long space (2).

We could maybe modify the regexp but this has caused troubles in the past: https://github.com/ledger/ledger-mode/issues/141.

ledger-post.el
DamienCassou: This is definitely not my best code :-). This code is here to handle the trailing comment if there is one: I added a comment.
Fortunately, we have tests to make sure the behavior is correct.

I will do whatever you suggest but I think asking the user to cleanup whitespace before being able to use the command isn't really nice to her.

DamienCassou:
I answered your comments. What do you think is the best way forward?

ledger-post.el
DamienCassou: @bcc32 I totally understand your point now, sorry. I added a commit simplifying the code and changing the test accordingly. If everything is fine, feel free to squash all these commits together.

DamienCassou:
I took care of your feedback. Can you please have a look at the last commit? Thank you so much for your guidance and time.

bcc32 (approved):

bcc32:
Thanks for your contribution, @DamienCassou!  Looks great, merged.

DamienCassou:
@bcc32 thank you so much for all your help. Most of the code is from you.

PR #422, opened 2024-04-11
Fix some bugs in ledger-complete-at-point date completion

### LINKED ISSUE #419: Wrong number of arguments calling ledger-complete-at-point on date
Example error:

```
Wrong number of arguments: (((dates (26140 42560) (25658 8512)) (now 26130 61652 191446 0)) (_string _predicate _all) (list (concat (ledger-format-date (cl-find-if #'(lambda (date) (not (time-less-p now date))) dates)) (and (= (point) (line-end-position)) " ")))), 0
```

The expected behavior should be that various dates (same day number in current
month, or previous month) are offered for completion.

bcc32:
ledger-complete-at-point currently seems completely broken for date completion
(e.g., "19" completes to the 19th of the current month, or the previous month,
whichever date is in the past from today).

I added some regression tests that demonstrate the bugs, and then separate
commits to fix each bug.  I plan to squash the test and fix commits together
when merging.

(@purcell, I spent some time fighting with the CI over some old calling
conventions for functions (func-arity, which didn't exist; and encode-time,
which doesn't support a single argument in Emacs < 27, but the multiple-argument
form is marked "obsolete" in Emacs 29).  What's ledger-mode's policy on
supporting Emacs major versions, is there generally some guideline we follow as
to when we're okay with abandoning support for older versions?)

Fix #419.

### DIFF
--- a/ledger-complete.el
+++ b/ledger-complete.el
@@ -240,9 +240,14 @@ an alist (ACCOUNT-ELEMENT . NODE)."
                       (if month
                           (encode-time 0 0 0 day month last-year)
                         (encode-time 0 0 0 day last-month last-month-year)))))
-    (list (concat (ledger-format-date
-                   (cl-find-if (lambda (date) (not (time-less-p now date))) dates))
-                  (and (= (point) (line-end-position)) " ")))))
+    (let ((collection
+           (list (concat (ledger-format-date
+                          (cl-find-if (lambda (date) (not (time-less-p now date))) dates))
+                         (and (= (point) (line-end-position)) " ")))))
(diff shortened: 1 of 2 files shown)


purcell:
> What's ledger-mode's policy on
> supporting Emacs major versions, is there generally some guideline we follow as
> to when we're okay with abandoning support for older versions?

I don't think it would be controversial to only support the latest three major versions, ie. dropping support for Emacs <= 26. I don't know if anyone is likely to declare a policy around this though.

purcell (approved):

PR #425, opened 2024-04-16
Attempt to fix flaky ledger-fontify/test-017 on Emacs 25.1

bcc32:
I'm not sure why this fix would be necessary, but it seems to make the CI happy.
(I would think `ledger-fontify-extend-region` should be sufficient to make
fontification work reliably, but it's possible Emacs 25.1 is doing something
differently).

### DIFF
--- a/test/test-helper.el
+++ b/test/test-helper.el
@@ -102,6 +102,10 @@ The two arguments START and END are character positions."
       (ledger-mode)
       (insert str)
       (font-lock-ensure)
+      ;; This second call appears to prevent `ledger-fontify/test-017' from
+      ;; nondeterministically failing on Emacs 25.1.  When ledger-mode no longer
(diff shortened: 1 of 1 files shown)


bcc32:
Guess I jinxed it.  CI passed on my fork but not here, so it's definitely something fishy going on.

purcell:
Since it's in a test about fontification, perhaps it would help to somehow force a font lock update, or disable jit-lock-mode... I haven't really dug into this code, sorry.

bcc32:
Well, it seems just calling `font-lock-ensure` a second time prevents the failures.  At least, I re-ran the Emacs 25.1 job five times by hand, and it succeeded each time.  Maybe good enough for now?

purcell (approved):

PR #428, opened 2024-06-06
When visiting report links, customize whether to go to xact beginning

bcc32:


bcc32:
@purcell, pretty straightforward code here but just checking to see if you think adding a customization to do this is reasonable.  (My personal use case is to be able to jump immediately to an uncleared posting from the report buffer and toggle it but not the whole xact)

PR #432, opened 2024-07-06
Adopt makem.sh for build support (linting, running tests)

### LINKED ISSUE #423: CI checkdoc step is just a no-op
test.yml specifies a checkdoc step (whose errors are currently suppressed), but
at the moment it doesn't actually do anything (and fails silently) because
https://download.tuxfamily.org/user42/checkdoc-batch.el is a broken link.

bcc32:
This PR switches from CMake and custom make rules to using
[makem.sh], as suggested in #423.

It also includes a small tweak to just disable `ledger-mode/test-001` (which
requires running Emacs interactively) when makem.sh runs tests, since it does
not provide a custom ERT selector.  That change can be removed once #431 is
merged.

Fix #423.

[makem.sh]: https://github.com/alphapapa/makem.sh


bcc32:
Discovered some issues on macos, opening alphapapa/makem.sh#49

bcc32:
Updated makem.sh to include the macOS fix.

purcell:
I'd be generally in favour of this, because I'm not sure if any of the more active maintainers have any specific love for cmake, and it would be good to include more linting etc. in the CI.

purcell:
> Isn't this just replacing one big blob (well a shell script isn't technically a blob but this one is very big) with another? Compared to that a Makefile is much shorter.

I'm a big Makefile fan, but the setup here is pretty overbuilt for current needs, misses common elisp build steps like linting, and none of the maintainers here are likely to want to add it. Meanwhile makem is fairly well used and actively supported, which is why I'm broadly in favour of the change. No other maintainers have expressed a preference either way.

> If this adapted I would recommend a submodule instead copying the file into the repository to make it easier to track the script.

That would be fine too, though I don't know if the script has to end up in the root of the repo to work properly.

Thaodan:
Steve Purcell ***@***.***> writes:

> purcell left a comment (ledger/ledger-mode#432)
>
>> Isn't this just replacing one big blob (well a shell script isn't
>> technically a blob but this one is very big) with another? Compared
>> to that a Makefile is much shorter.
>
> I'm a big Makefile fan, but the setup here is pretty overbuilt for
> current needs, misses common elisp build steps like linting, and none
> of the maintainers here are likely to want to add it. Meanwhile makem
> is fairly well used and actively supported, which is why I'm broadly
> in favour of the change. No other maintainers have expressed a
> preference either way.

Personally I hope that the classical

make
make install

still works.

CMake can handle test cases just fine but I get why it might not be the
right tool when there's no direct CMake Elisp package implementing all
the features.  

>> If this adapted I would recommend a submodule instead copying the file into the repository to make it easier to track the script.
>
> That would be fine too, though I don't know if the script has to end up in the root of the repo to work properly.

A symlink could work.

purcell:
> Personally I hope that the classical `make; make install` still works.

Seems very reasonable!

bcc32:
> > If this adapted I would recommend a submodule instead copying the file into the repository to make it easier to track the script. That would be fine too, though I don't know if the script has to end up in the root of the repo to work properly.
> A symlink could work.

Symlinks don't work reliably on Windows, AFAIK.

In the year since I first created this PR, no new release of makem.sh has occurred, and it's unlikely to change very frequently.  And, indeed the upstream repo recommends just copying the script into your project instead (whereas adding it as a submodule would force downloaders of ledger-mode to also download makem.sh's tests and documentation images).

(Note, also, that submodules are not included in `git clone` by default---meaning, users would have to take some special steps with fairly advanced git commands just so they can run a standalone bash script.)

As a result, IMO, it's not worth the hassle to try to use a submodule.

bcc32:
> I don't think that's a fair argument. Packages can come from distribution
> packages, it's a common practice.

On the other hand, it's very uncommon practice for Elisp packages to
provide their own installation command, though (I have never seen a
package do this other than mu4e, which is an extraordinary cas c
because it must be installed with the mu program at the same version).

If a distribution has a standard way of packaging and installing Elisp
packages, it's should just byte-compile, generate autoloads, etc. the
way it usually does.  I don't think ledger-mode devs have any special
insights on the right way to prepare such a package for every
platform.

> Further package-install-file depends on package.el while site-lisp
> just works with any Emacs package manager.

Yes, that's true of `package-install-file` specifically, but surely
other package managers can install ledger-mode from a checkout, or
from a tarball off of MELPA, no?  (Genuine question, since I have
never used an Emacs package manager other than package.el).  My point
is that those package managers may have some steps they want to take
in addition to the simple copying of files to some hardcoded path, and
we shouldn't second guess them.  It does not appear to be typical
practice in the ecosystem of Elisp packages I have seen.

PR #440, opened 2024-10-07
Hold on to ledger-reports variable's local value

errge:
When ledger-reports is defined as a file local variable, we have to inherit that local setting to the report buffer.

### DIFF
--- a/ledger-report.el
+++ b/ledger-report.el
@@ -291,10 +291,13 @@ used to generate the buffer, navigating the buffer, etc."
            (edit (not (null current-prefix-arg))))
        (list rname edit))))
   (let* ((file (ledger-master-file))
-         (buf (find-file-noselect file)))
+         (buf (find-file-noselect file))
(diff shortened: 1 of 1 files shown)


ledger-report.el
```diff
@@ -291,10 +291,13 @@ used to generate the buffer, navigating the buffer, etc."
            (edit (not (null current-prefix-arg))))
        (list rname edit))))
   (let* ((file (ledger-master-file))
-         (buf (find-file-noselect file)))
+         (buf (find-file-noselect file))
+         (local-ledger-reports (when (local-variable-p 'ledger-reports) ledger-reports)))
```
bcc32, line 295: I think it would be simpler to unconditionally use whatever value `ledger-reports` has in the current buffer.

bcc32 (changes requested):
Sorry for the long delay, I haven't had time to look at this project in a few months but am hoping to spend some time helping get it into shape for a new release.

Do you mind adding a regression test for this issue?

Issue #279, opened 2021-01-02
Pass --width to hledger when ledger-report-auto-width is t


DamienCassou:
I tried to make ledger-mode pass `--width` for hledger instead of `--columns` when `ledger-report-auto-width` is non-nil. I failed because of the way ledger-mode passes arguments to ledger/hledger:

- ledger-mode passes the `--columns` arguments right after the binary name: i.e., `ledger --columns 80 ...`
- hledger expects the corresponding `--width` [argument after the `register` command](https://hledger.org/hledger.html#custom-register-output): i.e., `hledger register --width 80 ...`

I don't know how to make that work without significantly rewriting this part of `ledger-mode`. Relevant functions include: `ledger-report--compute-extra-args`, `ledger-report-expand-format-specifiers` and `ledger-report-binary-format-specifier`.

simonmichael:
Related to #319.

PR #441, opened 2024-10-31
Added check to ignore --date-format option for hledger users

GLakovnik:
This PR intends to improve ledger-mode compatibility for hledger users.
It requires no additional configuration and is completely invisible to ledger users while fixing the most common error hledger users see.

It's "the dumbest thing that works", a simple check whether `ledger-binary-path` points to hledger. If it does, `ledger-exec-ledger` does not include the --date-format option (which hledger doesn't have).

Solves the largest problem for hledger users and is a prerequisite for other improvements (WIP, i plan to make reconcilliation work as well and this is the first step.)
I've documented the incompatibilities in [hledger's documentation](https://hledger.org/editors.html#editor-configuration).

With this change the following now works with hledger:
Quick balance display (\<C-c\> \<C-p\>)
Add transaction (\<C-c\> \<C-a\>)
Display ledger stats (\<C-c\> \<C-l\>)

### DIFF
--- a/ledger-exec.el
+++ b/ledger-exec.el
@@ -89,7 +89,7 @@ otherwise the error output is displayed and an error is raised."
                    (apply #'call-process-region
                           (append (list (point-min) (point-max)
                                         ledger-binary-path nil (list outbuf errfile) nil "-f" "-")
-                                  (when (not (string-match "hledger" ledger-binary-path))
+                                  (when ledger-default-date-format
                                     (list "--date-format" ledger-default-date-format))
                                   args)))))
             (if (ledger-exec-success-p exit-code outbuf)
--- a/ledger-exec.el
+++ b/ledger-exec.el
@@ -89,7 +89,8 @@ otherwise the error output is displayed and an error is raised."
                    (apply #'call-process-region
                           (append (list (point-min) (point-max)
                                         ledger-binary-path nil (list outbuf errfile) nil "-f" "-")
-                                  (list "--date-format" ledger-default-date-format)
+                                  (when (not (string-match "hledger" ledger-binary-path))
+                                    (list "--date-format" ledger-default-date-format))
                                   args)))))
             (if (ledger-exec-success-p exit-code outbuf)
                 outbuf
--- a/ledger-init.el
+++ b/ledger-init.el
@@ -59,11 +59,13 @@ Returns the current date if DATE is nil or not supplied.
 
 If FORMAT is provided, use that as the date format.  Otherwise,
 use the --input-date-format specified in `ledger-init-file-name',
-or if none, use `ledger-default-date-format'."
+or if none, use `ledger-default-date-format'. If none of
+the previous values exist, use ledger-iso-date-format."
   (format-time-string
    (or format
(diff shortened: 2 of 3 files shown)


bcc32:
Does hledger really not support setting a format for dates?  I'd rather do something different than just "silently don't respect the customized value of `--date-format`" in that case---perhaps we could support setting `ledger-default-date-format` to nil, which hledger users would have to do?

GLakovnik:
Sorry for how verbose this is -- hopefully this PR will give me enough clarity to submit further changes.

Hledger infers most formatting settings from input data -- transactions and rules. AFAIK hledger doesn't have a specific setting for date-format. It defaults to ISO dates but adjusts its output based on input data and supports at least 2000-01-01, 2000/01/01, 2000.01.01, which it calls simple dates. While it does have a date-format option, it's a helper to define a format string for parsing dates from CSV files specifically. The allowed date formats in transaction data and program output is limited to the simple dates, while date definitions in commandline options like --begin, --end and its ilk use 'smart dates', allowing for strings like 'this month' and similar conveniences.

  In a functional sense nothing is lost for hledger users by ignoring --date-format. The only use-case i can see is if someone would want to write their transaction data in one format but have hledger output data in another... which sounds strange.
  
  In regards to setting ledger-default-date-format to nil:
  If the only function of ledger-default-date is to pass its value to ledger/hledger then setting it to nil is fine. With a quick grep through the code i didn't find other uses currently, but i could imagine a world where that value could be used in other contexts in the future -- say if ledger-mode used it to generate a transaction template, for example. This might be overthinking it tho, i think setting it to nil is fine.

  This PR would then simply use a different condition:
```elisp
(when ledger-default-date-format
  (list "--date-format" ledger-default-date-format))
```

I do think that my proposed solution is a bit 'ugly' and this version is prettier.

#### The following relates to other incompatibilities and possible future changes

  The PR was mainly intended to get the ball rolling -- i wanted to propose something and see what the community/maintainer reaction would be. My goal here is to understand the code style expected so i can submit further PRs and reduce friction. Most incompatibility can be fixed by adjusting flags for ledger/hledger invocation. As an example, an incompatibility that got resolved in the past was the differences in color settings.
  Ledger uses '--color --force-color', hledger uses '--color=always'.
This got resolved by creating the custom variable ledger-report-native-highlighting-arguments.
There are other similar differences which hledger users currently solve with a shell script that modifies command-line options before calling hledger, namely:
  - changing 'xact' to 'print --match'
  - changing 'cleared' to 'balance -N -C'
There are other equivalences related to reconciliation, but the usefulness of those is dependent on work i haven't done yet.
All of these along with a compatibility checklist are documented in hledgers documentation.

Can i use the 'introduce a custom variable' method for these differences as well?
Ideally i would like to submit 3 or 4 quick PRs to make these changes and then tackle reconcilliation -- i don't know what that entails yet.

#279 -- Relevant question by another hledger user about another one of these minor differences.

Thank you for your time reading all this, i hope to cause less distraction in the future.

Thaodan:
Could you rebase this PR instead of merging master?
Maybe also squash some of the commits.

PR #446, opened 2025-03-13
CI: Fix usage of soon-to-be obsolete items

bcc32:
if-let, when-let, and gensym are obsolete starting in Emacs 31.1.  This commit
fixes the Emacs snapshot CI build, but unfortunately needs to drop support for
Emacs 25.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -15,17 +15,11 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest]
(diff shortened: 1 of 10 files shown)


bcc32:
@purcell, do you think it is reasonable to drop support for Emacs 25?  Emacs 26 was released in 2018.  I suppose we could also wait until Emacs 31 is actually released before we need to do this, but it would be a little unfortunate to have the snapshot CI continually failing until then.  (Is it possible to suppress the snapshot CI failure emails?)

Issue #450, opened 2025-08-10
Ledger won't visit the correct place in source file from the Ledger Check buffer


enderw88:
Using the freshest ELPA version, run the ledger check function from the menu.  You should be able to hit <RET> and have it go to that place in the buffer.  If will go nearly to the beginning of the buffer

This is because the Ledger Check report uses EMACS markers while the visit source code expect line numbers, which is what the various ledger reports will put out.

I have a fix ready, but i have just move to a new machine and haven't gotten my GITHUB authentication working yet.

PR #452, opened 2025-08-10
Fix visit source

enderw88:
Added logic to handle whether the report buffer was using emacs markers or line number to visit source.  Fixes Issue #450


purcell:
Current CI run successful, so merging.

PR #455, opened 2025-09-03
Add comments completion

immae:
There used to be a bug in ledger mode which allowed comments (including tags & meta) to get completed because they were somehow wrongly parsed as an account. This bug was very useful to me since I rely a lot on tags in my accounting, and its fixing made my work much harder.

This change is an attempt to re-add the completion feature (disclaimer: the code was bootstrapped with an LLM but failed to produce any usable result; however it helped me understand how lisp works and find relevant functions, so most of the code is still written by me)

As is, the code does the job, but it contains a few issues (it’s still a win as is in my opinion):

- The listing doesn’t get cached (not sure it is a problem, the completion is very fast for me)
- The listing is incomplete: it only lists account comments, not comments that appear on the same line after payees or prices (but completion still works in these positions)
- The completion itself could be improved: it completes immediately with a whole string instead of completing until biggest common
- The implementation of the search-forward should probably rely on `ledger-comment-string` but the regexp seems incorrect as is

/cc @simonmichael

### DIFF
--- a/ledger-complete.el
+++ b/ledger-complete.el
@@ -308,8 +308,6 @@ an alist (ACCOUNT-ELEMENT . NODE)."
            (eq 'comment (car (cdr (ledger-context-at-point))))
            (save-excursion
              (back-to-indentation)
-             (search-forward ";")
-             (skip-chars-forward " \t")
              (setq start (point)))
            (setq collection (cons 'nullary #'ledger-comments-list)))
           (;; Payees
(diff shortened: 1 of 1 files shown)


immae:
@bcc32 I just added your changes don’t see any difference with what I had before, so I guess it’s a good sign? I think I understand what you changed and it seems fine by me anyway!

Thanks for your improvement and the tests!

PR #459, opened 2025-12-05
Clean up and modernize ledger-check

bcc32:
1. Implement missing redo command
1. Correctly read input from current buffer like ledger-check-buffer's docstring
   says
1. Improve window handling logic
1. Derive major mode from `special-mode`


bcc32:
~~Now depends on #474 as well.~~

PR #461, opened 2025-12-08
Add support for live preview of ledger xact output while adding transactions

### LINKED ISSUE #456: ledger-add-transaction candidate narrowing
Hello!

I usually add a new transaction by interactively calling `ledger-add-transaction`, which prompts for a date and then prompts for a `xact` string. The string prompt has become a pain point as I have a ton of past payees in my file and it's really hard to guess what payee ledger will guess from a given string.

Is there any way to preview/narrow the existing payees in the minibuffer when using xact?

bcc32:
Add a new customization, `ledger-add-transaction-idle-preview` (enabled by
default) to display a preview of the `ledger xact` output that would be inserted
in the ledger buffer as the user is inputting a transaction text in
`ledger-add-transaction`.  (Also affects `ledger-reconcile-add`).

`ledger xact`'s exact output can be quite unpredictable, since it is hard to
remember exactly what payees exist in a ledger file, and since it depends on the
most recent transactions that match the inputted strings.  Displaying a live
preview that can inform the user how to make their input more specific is a much
nicer experience than "hit enter, see that the resulting transaction is wrong,
undo, press C-c C-a again, enter date again, then press M-p to bring up previous
input".

Close #456.


bcc32:
I've fixed the bad interactions with while-no-input by switching to
make-process.  This caused a performance hit, which I fixed by writing the
buffer contents to a temp file, which is an improvement over call-process-region
anyway.

bcc32:
Currently blocked on #474.

bcc32:
Unfortunately, it seems necessary to decrease the test coverage requirement due to this code all being effectively untestable, at least not robustly.

On the other hand, I have been running this branch of ledger-mode myself for the last 6 months, and it works smoothly.

PR #467, opened 2026-04-24
Respect year directive when sorting transactions

### LINKED ISSUE #343: Take ledger-default-date-format in to account for sorting
`SPC m C (ledger-mode-clean-buffer)` doesn't sort as expected in case of a non-alphabetically sorting date format.  
Configuring my default date format doesn't change anything to this.
```elisp
(defun dotspacemacs/user-config ()
  "Configuration for user code:
This function is called at the very end of Spacemacs startup, after layer
configuration.
Put your configuration code here, except for variables that should be set
before packages are loaded."
  (setq ledger-default-date-format "%d/%m/%Y")
  (setq ledger-init-file-name "/home/michael/.ledgerrc")
)
```
This might be due to https://github.com/ledger/ledger/issues/1068 , but then the wording `sort chronologically` [in ledger-mode sources](https://github.com/ledger/ledger-mode/blob/1e6ddafa0fdf8f009c35488fdd6d0b0c8485ad34/ledger-sort.el#L66) is at least a little misleading.

### LINKED ISSUE #1068: ledger-sort-region and ledger-sort-buffer should actually parse dates (BZ#1068)
Note: the issue was created automatically with bugzilla2github

Bugzilla bug ID: BZ#1068
From: Eric Abrahamsen (@girzel)
CC:   unknown user

jwiegley:
## Summary

- `ledger-sort-startkey' now parses the date at point via `ledger-iso-date-regexp' and correctly handles non-padded dates (e.g. `2015/5/7').
- Dates written without a year component (e.g. `07/17') are interpreted relative to the most recent `year NNNN' or `Y NNNN' directive preceding the transaction, looking through any active narrowing so sorting a sub-region still sees the enclosing directive.  The current calendar year is used when no directive is in scope.
- Previously such short dates fell through to `(float-time nil)', which returns "now", so they sorted as today's date regardless of the surrounding `year' context.

Fixes ledger/ledger#1068.

## Test plan

- [x] `make test` in `test/` passes (241 tests).
- [x] New ERT tests in `test/sort-test.el` cover:
  - `year NNNN' directive applied to year-less dates.
  - `Y NNNN' short form of the directive.
  - Non-padded dates (`2015/5/7' sorts before `2015/5/10').
  - `year' directive above a narrowed sort region.

### DIFF
--- a/ledger-sort.el
+++ b/ledger-sort.el
@@ -59,13 +59,41 @@
   (beginning-of-line)
   (insert "\n; Ledger-mode: End sort\n\n"))
 
+(defconst ledger-sort--year-directive-regex
+  "^\\(?:Y\\|year\\)\\s-+\\([0-9]+\\)\\s-*$"
+  "Regex matching a `year NNNN' or `Y NNNN' directive at the start of a line.
+The year number is captured in group 1.")
+
+(defun ledger-sort--preceding-year ()
+  "Return the year from the most recent `year NNNN' or `Y NNNN' directive.
+Searches backward from the start of the current line, ignoring any
+restriction so that directives above a narrowed sort region are still
+consulted.  Returns a number, or nil if no such directive exists."
+  (save-excursion
+    (save-restriction
+      (widen)
+      (beginning-of-line)
+      (when (re-search-backward ledger-sort--year-directive-regex nil t)
+        (string-to-number (match-string 1))))))
+
 (defun ledger-sort-startkey ()
-  "Return a numeric sort key based on the date of the xact beginning at point."
+  "Return a numeric sort key based on the date of the xact beginning at point.
+Dates with a full four-digit year are parsed directly.  Short dates of the
+form M/D or MM/DD are interpreted relative to the most recent `year NNNN'
+directive preceding the current transaction, falling back to the current
+calendar year if no such directive exists."
   ;; Can use `time-convert' to return an integer instead of a floating-point
   ;; number, starting in Emacs 27.
   (float-time
-   (ledger-parse-iso-date
-    (buffer-substring-no-properties (point) (+ 10 (point))))))
+   (cond
+    ((looking-at ledger-iso-date-regexp)
+     (ledger-parse-iso-date (match-string 0)))
+    ((looking-at "\\([0-9]\\{1,2\\}\\)[-/]\\([0-9]\\{1,2\\}\\)\\(?:[^-/0-9]\\|$\\)")
+     (let ((month (string-to-number (match-string 1)))
+           (day   (string-to-number (match-string 2)))
+           (year  (or (ledger-sort--preceding-year)
+                      (nth 5 (decode-time)))))
+       (encode-time 0 0 0 day month year))))))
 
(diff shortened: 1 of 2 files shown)


ledger-sort.el
```diff
@@ -59,13 +59,41 @@
   (beginning-of-line)
   (insert "\n; Ledger-mode: End sort\n\n"))
 
+(defconst ledger-sort--year-directive-regex
```
bcc32, line 62: This should probably go in `ledger-regex.el` for consistency.

ledger-sort.el
```diff
@@ -59,13 +59,41 @@
...
+  "^\\(?:Y\\|year\\)\\s-+\\([0-9]+\\)\\s-*$"
+  "Regex matching a `year NNNN' or `Y NNNN' directive at the start of a line.
+The year number is captured in group 1.")
+
+(defun ledger-sort--preceding-year ()
```
bcc32, line 67: This algorithm is very quadratic.  On my personal ledger file with 13k transactions (3 MB file), if I convert it to use year directives instead of YYYY-MM-DD, `ledger-sort-buffer` takes 6-7 seconds just gathering sort keys.

Perhaps a better approach would be to, before sorting, scan once from the beginning of the buffer until the end of the sort region for year directives, and look up the year for a transaction based on its position in the buffer relative to the year directives found (which ought to be a pretty small number).

ledger-sort.el
```diff
@@ -59,13 +59,41 @@
...
-   (ledger-parse-iso-date
-    (buffer-substring-no-properties (point) (+ 10 (point))))))
+   (cond
+    ((looking-at ledger-iso-date-regexp)
+     (ledger-parse-iso-date (match-string 0)))
+    ((looking-at "\\([0-9]\\{1,2\\}\\)[-/]\\([0-9]\\{1,2\\}\\)\\(?:[^-/0-9]\\|$\\)")
+     (let ((month (string-to-number (match-string 1)))
+           (day   (string-to-number (match-string 2)))
+           (year  (or (ledger-sort--preceding-year)
+                      (nth 5 (decode-time)))))
```
bcc32, line 95: It seems very strange for this function to depend on the actual current wall time.  Is that compatible with the actual semantics of ledger here?  (admittedly I do not personally use year directives)

ledger-sort.el
```diff
@@ -59,13 +59,41 @@
...
+calendar year if no such directive exists."
   ;; Can use `time-convert' to return an integer instead of a floating-point
   ;; number, starting in Emacs 27.
   (float-time
-   (ledger-parse-iso-date
-    (buffer-substring-no-properties (point) (+ 10 (point))))))
+   (cond
+    ((looking-at ledger-iso-date-regexp)
+     (ledger-parse-iso-date (match-string 0)))
+    ((looking-at "\\([0-9]\\{1,2\\}\\)[-/]\\([0-9]\\{1,2\\}\\)\\(?:[^-/0-9]\\|$\\)")
```
bcc32, line 91: I think this is more-or-less equivalent to ledger-incomplete-date-regexp.

PR #477, opened 2026-06-10
Replace ledger-get-minibuffer-prompt with built-in format-prompt

bcc32:
This introduces a dependency on the `compat` package.  Not sure if the other
maintainers think that might be controversial, but doing so would allow many
other things to take advantage of more modern APIs, and faster, while retaining
compatibility with older Emacs versions.

### DIFF
--- a/tools/lint.el
+++ b/tools/lint.el
@@ -18,6 +18,10 @@
 
 ;;; Code:
 
+;; Required for `package-lint' to find installed `compat' package and not warn
+;; about the dependency.
+(package-activate-all)
+
 (require 'package-lint nil t)
 (require 'checkdoc)
 
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -45,10 +45,13 @@ jobs:
     - uses: purcell/setup-emacs@master
       with:
         version: ${{ matrix.emacs_version }}
+    - name: Install compat
+      run: emacs --batch -f package-initialize -f package-refresh-contents --eval "(package-install 'compat)"
 
     - uses: actions/checkout@v6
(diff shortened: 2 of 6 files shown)


tools/lint.el
```diff
@@ -18,6 +18,10 @@
 
 ;;; Code:
 
+;; Required for `package-lint' to find installed `compat' package and not warn
+;; about the dependency.
+(package-initialize)
```
basil-conto: Would `package-activate-all` work here instead? If so, it's probably the more suitable choice.

bcc32: Seems so, thanks for the suggestion

bcc32:
(requesting some review due to the introduction of the compat dependency)

tools/lint.el
```diff
@@ -18,6 +18,10 @@
 
 ;;; Code:
 
+;; Required for `package-lint' to find installed `compat' package and not warn
```
purcell, line 21: `package-lint` doesn't need the dependent packages to be installed, it just needs to know they are installable.

purcell:
I'm a bit on the fence about this. On the one hand it would likely be fine to raise the minimum Emacs version required by `ledger-mode` to 27.x, or _perhaps_ even 28. Requiring extra packages is slightly more drastic, and while I'm happy `compat` exists, I don't think it always provides perfect compatibility. I'm not hugely opposed though — depends a bit on what improvements it could yield here.

bcc32:
Fair enough.  Obviously, `format-prompt` isn't a compelling enough reason.  I'll drop this for now, and we could potentially revisit introducing `compat` if more things come up that it could help with.

